PR #1, opened 2020-04-06
Basic joystick teleop app for the tractor.

ethanrublee:
- Adds a python module called farm_ng.
- Add a simple canbus driver
  based off of https://github.com/abencz/python_socketcan, MIT licensed)
- Add a simple linux joystick drive
  based off of https://gist.githubusercontent.com/rdb/8864666/raw/516178252bbe1cfe8067145b11223ee54c5d9698/js_linux.py

This needs some clean up, but is working on the robot as seen in this  (https://youtu.be/hc9ApQ0mXlU)

### DIFF
--- a/python/farm_ng/joystick_teleop.py
+++ b/python/farm_ng/joystick_teleop.py
@@ -1,0 +1,92 @@
+from farm_ng.canbus import CANSocket, format_data
+import socket
+from farm_ng.joystick import Joystick
+import sys
+import select
+import linuxfd
+import time
+import math
+import struct
+
+VESC_SET_DUTY=0
+VESC_SET_CURRENT=1
+VESC_SET_CURRENT_BRAKE=2
+VESC_SET_RPM=3
+VESC_SET_POS=4
+        
+def send_rpm_command(canbus, motor_id, rpm):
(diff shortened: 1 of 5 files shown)


setup.bash
```diff
@@ -0,0 +1,11 @@
+# Get directory where this script is located
```
mikeferguson, line 1: see, you already know like half of ROS - source your setup.bash to get the paths right!

python/farm_ng/joystick_teleop.py
```diff
@@ -0,0 +1,92 @@
...
+import socket
+from farm_ng.joystick import Joystick
+import sys
+import select
+import linuxfd
+import time
+import math
+import struct
+
+VESC_SET_DUTY=0
```
mikeferguson, line 11: so I know this is just a quick demo, but at some point we probably should abstract out the motor driver instance to a class, which has a well defined API -- that would allow us (or anyone) to swap out the motor drivers a lot easier

ethanrublee: Yeah that makes sense,  I'll do another pass to add interfaces as this matures.

Issue #7, opened 2020-04-22
driver service


ethanrublee:
Add a low level service to the driver with access to all the hw required for driving the tractor

This service will be responsible for communicating with:

- joystick
- motor drivers over canbus
- camera

And will optionally depend on the rtkrover service.

Its roll will be to execute teleop and path following, with basic obstacle avoidance.

Will expose a remote api to the frontend service TBD.

PR #9, opened 2020-04-22
Driver service

ethanrublee:
This adds the basic canbus + joystick driver to the docker-compose based app, #7 

- Survives reboots, tested hard and soft
- Joystick can be plugged and uplugged
- docker starts socket can, sets the gpio pins with busybox devmem, and starts up the bluetooth
- Removes the app/tractor/canbus and replaces it with app/tractor/driver
- Moved the farm_ng.joystick_teleop package to farm_ng.tractor.driver

Pairing of the joystick is still done manually.


docs/app.md
```diff
@@ -38,13 +38,13 @@ farmer@tractor01:~$ ~/code/tractor/env.sh
 To build the tractor app on the tractor/xavier (TODO(rublee) make workflow from workstation with docker-compose or balena):
```
mikeferguson, line 38: is this comment now outdated?

python/farm_ng/joystick.py
```diff
@@ -9,174 +9,213 @@
...
+
+        if type & 0x02:
+            axis = self.axis_map[number]
+            if axis:
+                fvalue = value / 32767.0
+                self.axis_states[axis] = fvalue
+                # print("%s: %.3f" % (axis, fvalue))
+
+
+class MaybeJoystick:
```
mikeferguson, line 175: maybe add a comment on what this class is for (and why someone would use MaybeJoystick instead of Joystick directly)

docs/app.md
ethanrublee: No, at the moment I've only figured out how to use docker-compose on the xavier.  I'd like to figure out deploy from my workstation, where it builds on my workstation, and does a push somehow to the robot.  Balena seems promissing but I haven't had the chance to dig into why bluetooth doesn't come up (https://forums.balena.io/t/nvidia-xavier-bluetooth-support/96532)

python/farm_ng/joystick.py
ethanrublee: Thanks, added basic comment, yeah need to figure out how to enforce documentation... TODO :)

Issue #10, opened 2020-04-22
code formatting githook


ethanrublee:
Add some githooks for local dev and as part of the github pull request checks.

Python: maybe black https://github.com/psf/black, yapf, ...
Dockerfile: not sure here, found this https://github.com/replicatedhq/dockerfilelint
Not much c++ yet, but soon, clang-format

PR #15, opened 2020-04-26
Adding pre-commit config.

ethanrublee:
Adds pre-commit hooks, see issue #10


ethanrublee:
Also add some development docs here: https://github.com/farm-ng/tractor/wiki/Development-guide

PR #20, opened 2020-05-17
Add structure for frontend development

isherman:
Including:
- yarn (dependency management)
- eslint (linting)
- prettier (auto-formatting)
- vscode (IDE integration)
- babel (compilation)
- webpack (bundling)
- git hook

Ported tractor visualization code from `ethanrublee:rtk-record`


app/frontend/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "eslint-plugin-prettier": "^3.1.3",
+    "eslint-plugin-react": "^7.20.0",
+    "prettier": "^2.0.5",
+    "style-loader": "^1.2.1",
+    "webpack": "^4.43.0",
+    "webpack-cli": "^3.3.11",
+    "webpack-dev-server": "^3.11.0"
+  },
+  "dependencies": {
+    "three": "^0.116.1"
```
ethanrublee, line 28: Cool so does this install the threejs from some upstream repository?

app/frontend/README.md
```diff
@@ -0,0 +1,18 @@
...
+- If using Visual Studio Code, `.vscode/` configuration will handle linting and auto-format on save.
+
+### Install dependencies
+```
+yarn
+```
+
+### Launch the web app
+```
+yarn dev-server
```
ethanrublee, line 16: How do we support the python backend here?  or are you doing it with some magic somehow.

app/frontend/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+### Prerequisites
+- Install [Yarn](https://classic.yarnpkg.com/en/docs/install)
+
+### Recommendations
+- If using Visual Studio Code, `.vscode/` configuration will handle linting and auto-format on save.
+
+### Install dependencies
+```
+yarn
```
ethanrublee, line 11: Are we supposed to commit the yarn.lock file?  Looks like i have a little to learn here, but looks cool.

app/frontend/src/index.js
```diff
@@ -0,0 +1,140 @@
...
+let scene;
+let renderer;
+let tractorMesh;
+let xAxis;
+
+const simUpdater = {
+  socket: null,
+
+  start() {
+    const url = "ws://localhost:8989/simsocket";
```
ethanrublee, line 16: Oh, so we can have a separate server running for the backend, other than the server/port hosting the frontend javascript files?

app/frontend/package.json
```diff
@@ -0,0 +1,34 @@
...
+  "license": "All Rights Reserved",
+  "private": true,
+  "devDependencies": {
+    "@babel/cli": "^7.8.4",
+    "@babel/core": "^7.9.6",
+    "@babel/preset-env": "^7.9.6",
+    "@babel/preset-react": "^7.9.4",
+    "babel-loader": "^8.1.0",
+    "css-loader": "^3.5.3",
+    "eslint": "^7.0.0",
```
ethanrublee, line 15: I'm seeing this error on my machine after running ``yarn``

```
error eslint@7.0.0: The engine "node" is incompatible with this module. Expected version "^10.12.0 || >=12.0.0". Got "8.10.0"
error Found incompatible module.
```

ethanrublee: I fixed this by upgrading node on my machine using:

```
sudo npm install -g n
sudo n stable
```

now i have
```
$ node --version
v12.16.3
```

app/frontend/package.json
isherman: Yeah, running `yarn` (short for `yarn install`) fetches these dependencies from the npm repository (recently acquired by Github). Javascript library authors tend to release to npm fairly often.

app/frontend/README.md
isherman: Yeah, the idea here is this is just a convenient little dev server when you're iterating on the frontend only. Should support auto-reload-on-save, that kind of thing.

`yarn build` outputs all the bundled up JS/HTML/CSS to `dist/`. We'd then use our build system to drop those files somewhere the Python server can see them (and serve them).

app/frontend/README.md
isherman: Yeah, convention is to check this in. You end up specifying the dependencies in `package.json` in a more general way, and then `yarn.lock` guarantees when we run `yarn install` on CI/CD or local developer machines we'll get the same version of all dependencies. Running `yarn upgrade` updates the versions in the `.lock` file to the latest versions available (within the range specified in `package.json`).

app/frontend/src/index.js
isherman: Right, that's the workflow I was using. We should be able to support both workflows.

app/frontend/package.json
isherman: Cool. I specified v12 on CI. Presumably developers could run the frontend dockerized to avoid this but I think there's some benefit to recommending a minimum Node version for anyone doing frontend development.

isherman:
Related to your question about running separate frontend and backend server. This did require me to make the following change in the Python server to avoid 403s on the websocket connection:

```python
--- a/python/farm_ng/tractor/sim/__main__.py
+++ b/python/farm_ng/tractor/sim/__main__.py
@@ -63,6 +63,9 @@ class SimSocketHandler(tornado.websocket.WebSocketHandler):
     def on_close(self):
         SimSocketHandler.waiters.remove(self)

+    def check_origin(self, origin):
+        return True
+
```

Obviously not code you'd want to ship to production. A wrinkle to iron out.

app/frontend/package.json
isherman: Part of a well-configured bundling process is that these upstream dependencies get "tree-shaken", so only the code paths in the dependency that are actually used end up in the bundle.

ethanrublee:
This looks like an example using the webpack and tornado (both watch for changes) for dev serving https://github.com/yychen/tornado-webpack

ethanrublee:
I'm going to try and get the rtk-record branch merged even if its a little half baked into master, so that I can test this more easily with the sim backend.

isherman:
> I'm going to try and get the rtk-record branch merged even if its a little half baked into master, so that I can test this more easily with the sim backend.

(I ended up cloning a second copy of

ethanrublee:
Ok, I merged the big sketchy branch with master. I'll test this more tomorrow.  Thanks for all the great stuff in here.

ethanrublee:
Sweet, after adding the check_origin hack to the sim server it works. Merging.

PR #22, opened 2020-05-23
Protobuf-generated clients & servers, Typescript, React

isherman:
Sorry for the large PR -- this is the sum of three different efforts. There's a lot to talk about. Happy to break it up however you'd like!

The first goal was to use React in the frontend, and represent the 3D scene declaratively with a library called `react-three-fiber`. You can see this in its own PR in my repository [here](https://github.com/isherman/tractor/pull/1).

The second goal was to use Typescript in the frontend. I can't say enough good things about TS. You can see this in its own PR in my repository [here](https://github.com/isherman/tractor/pull/2).

The third goal was to use protobufs for model and service definitions, and do a proof-of-concept for type-safe clients, automatic server-side validation, etc. from a single source of truth.

I tried a few things here, and ultimately ended up prototyping a solution based on [Twirp](https://github.com/twitchtv/twirp). What I like about Twirp is that you get both a JSON and a binary API, over HTTP 1.1 and HTTP/2, without the gRPC dependency or the need to run a HTTP/gRPC transcoding proxy (which is [how Google Cloud does it](https://github.com/cloudendpoints/esp)). It's just a very thin set of conventions on top of protobuf service definitions. It also seems to have a nice, but small, community contributing tools, client/server generators, etc. 

One thing I explored, which would have been relevant in my last project, was mapping rpcs onto a RESTful HTTP API via annotations in the service proto (again, how Google Cloud does it). But for a robot, I think a non-RESTful API is more appropriate, and then you don't have to worry about all the HTTP verbs, etc.

There are still a bunch of open questions about API semantics and protocols -- I've been trying to take some notes [here](https://github.com/isherman/tractor/wiki/Robot-API-and-Web-Clients) -- but this is an attempt to prototype an approach I've been wanting to try for a long time. Personally I'm excited about it!

Screen Shot 2020-05-22 at 5 05 25 PM

Screen Shot 2020-05-22 at 5 05 58 PM

It does require codegen of both client-side and server-side stubs. This should all be encapsulated by `make protos`. The server-side generator is a couple hundred lines of Go code [we'd maintain](https://github.com/isherman/protoc-gen-twirp_tornado_srv). The client-side generator is a [third-party library](https://github.com/stephenh/ts-proto) with a small but growing userbase.

Stuff in here:
- end-to-end binary unary rpcs (over HTTP) and server-streaming binary messages (over websocket)
- placeholder proto and service definitions, for testing
- protobuf conventions (e.g. wrappers for "nullable" fields)
- server-side validation (using `validate.rules` annotations in proto definition)
- an example service implementation
- some placeholder on-robot "models", to illustrate `fromProto` and `toProto` pattern
- protobuf linting (via [buf](https://github.com/bufbuild/buf))
- more vscode support
- python testing with pytest

There are a few TODOs scattered in here. I'd also like to:
- Add support for generating JSON clients (not just protobuf clients) to the upstream [ts-proto](https://github.com/stephenh/ts-proto) library.
- Integrate tests with CI/CD (I couldn't figure out how to launch a tornado server from my `pytest`)


ethanrublee:
Awesome, I'll have some time tomorrow to take a look.  Big PRs are OK,
especially since we're in the breadth first phase :)

Ethan

isherman:
> * Integrate tests with CI/CD (I couldn't figure out how to launch a tornado server from my `pytest`)

Ended up getting this working (as integration tests).

.vscode/extensions.json
```diff
@@ -1,6 +1,8 @@
 {
```
ethanrublee, line 1: I should get a tutorial on how to setup vscode for this project.

app/frontend/test/api.int.test.ts
```diff
@@ -0,0 +1,60 @@
...
+describe("WaypointService", () => {
+  const protobufClient = createTwirpClient(
+    WaypointServiceClientImpl,
+    "protobuf"
+  );
+
+  async function createWaypoint(): Promise<number> {
+    const response = await protobufClient.CreateWaypoint({
+      waypoint: {
+        lat: 42,
```
ethanrublee, line 16: For the v1, i'm leaning towards using ENU relative to an RTK base station, rather than Lat Long for waypoints, https://en.wikipedia.org/wiki/Local_tangent_plane_coordinates.  ENU relative to the base station, means that the system can turn on, and be accurate relatively quickly (< 5 minutes), rather than surveying the base station and getting accurate Lat, Long.  Also, on a small farm, I'm realizing that elevation may be important, so its convenient to just have a local 3d reference frame.  For integration with maps or satellite imagery, I'm not sure I have a good use case for this yet, but here, it would be useful to have lat/longs for waypoints.  The trouble is that lat/long might only be accurate to 2-4 meters unless the base station is surveyed.

Been thinking of modeling the farm as a Cartesian volume, if overhead map is desired, registering the Cartesian volume to Lat/Long through the base station, then if we improve base station localization, everything will get it.  So waypoint transform to lat long would be something like, lat_lon = enu_to_lat_lon(base_station.lat_lon, waypoint.enu)

protos/farmng/tractor/v1/geometry.proto
```diff
@@ -0,0 +1,39 @@
...
+import "google/protobuf/wrappers.proto";
+
+package farmng.tractor.v1;
+
+message Vec2 {
+  double x = 1;
+  double y = 2;
+}
+
+message PartialVec2 {
```
ethanrublee, line 12: Is this where either x/y is optionally set, and you have to test for the presence of the field?

protos/farmng/tractor/v1/status.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto3";
+
+import "farmng/tractor/v1/geometry.proto";
+
+package farmng.tractor.v1;
+
+message StreamStatusRequest {}
+
+message Status {
```
ethanrublee, line 9: Maybe this should be qualified with a better name, can be done later, just reading this and I know its inherited naming from my other prototype :)

python/farm_ng/tractor/sim/__main__.py
```diff
@@ -55,20 +66,24 @@ class SimSocketHandler(tornado.websocket.WebSocketHandler):
...
+        return None
 
     def open(self):
         SimSocketHandler.waiters.add(self)
 
     def on_close(self):
         SimSocketHandler.waiters.remove(self)
 
+    def check_origin(self, origin):
+        return True  # TODO: REMOVE
```
ethanrublee, line 79: I think it should be straight forward to support webpack and folder monitoring from tornado.  TODO in follow up.

python/farm_ng/tractor/sim/__main__.py
```diff
@@ -107,15 +122,31 @@ async def loop(self):
...
 
-            SimSocketHandler.send_updates(
-                dict(
-                    world_translation_tractor=self.world_pose_tractor.trans.tolist(),
-                    world_quaternion_tractor=self.world_pose_tractor.rot.to_quaternion(
-                        ordering='xyzw',
-                    ).tolist(),
-                    t=t,
+            status_proto = status_pb2.Status(
+                pose=geometry_pb2.SE3Pose(
```
ethanrublee, line 126: Would be nice to have an SE3 to/from proto convenience function.

python/farm_ng/tractor/sim/__main__.py
```diff
@@ -55,20 +66,24 @@ class SimSocketHandler(tornado.websocket.WebSocketHandler):
...
+    def check_origin(self, origin):
+        return True  # TODO: REMOVE
+
     @classmethod
     def send_updates(cls, status_msg):
         logger.debug('sending message to %d waiters', len(cls.waiters))
         for waiter in cls.waiters:
             try:
-                waiter.write_message(status_msg)
+                waiter.write_message(status_msg, binary=True)
```
ethanrublee, line 86: Binary, nice.

python/farm_ng/tractor/sim/models/waypoint.py
```diff
@@ -0,0 +1,47 @@
...
+@dataclass
+class Waypoint:
+    """
+    Models like this should exist on-robot, separate from their serialized representation.
+    Models provide helper functions to (de)serialize to/from the network, database, etc.
+    """
+    id: int
+    lat: float
+    lng: float
+    angle: float
```
ethanrublee, line 18: Is angle a desired heading at the goal?

python/farm_ng/tractor/sim/models/waypoint.py
```diff
@@ -0,0 +1,47 @@
+from dataclasses import dataclass
+from typing import Optional
+
+from farm_ng.tractor.sim.models.duration import Duration
+from farmng.tractor.v1 import waypoint_pb2
+from google.protobuf import wrappers_pb2
+
+
+@dataclass
+class Waypoint:
```
ethanrublee, line 10: So, we still need python models like this because they may enforce some class invariance that can't be expressed in the protobuf, or not convenient to use from python?

scripts/build-protos.sh
```diff
@@ -0,0 +1,18 @@
...
+CMD_ARGS="--proto_path=protos
+          --python_out=python/genproto
+          --ts_proto_out=app/frontend/genproto
+          --ts_proto_opt=forceLong=long
+          --twirp_tornado_srv_out=python/gensrv
+          protos/farmng/tractor/v1/*.proto
+          protos/validate/validate.proto"
+TAG="protoc"
+
+docker build -t $TAG -f $DIR/Dockerfile.protoc $DIR
```
ethanrublee, line 15: I like the mini-docker to build the protos.

On my last project, we used a monolithic dev docker, which has pros/cons, however we were more all C++ and a little python, maybe for mixture of go, javascript, python and c++ micro-dev dockers are easier to grapple with.

Haven't figured out the dev environment yet, but I like how this is evolving organically here.

.vscode/extensions.json
isherman: Should be as simple as `cd <root_of_your_repo>; code .`
Vscode will prompt you to install some suggested extensions (those listed in this file). 
This file might be missing a couple entries, but yeah the goal is a good default experience for vscode is tracked in version control -- would be great to get your feedback.

protos/farmng/tractor/v1/geometry.proto
isherman: Yeah, was hoping this would spark a discussion. In proto3 everything is optional, and the recipient has no way to distinguish between a "null" value for a field vs. its default value. `SetVelocity(Vec2{x=1, y=0})` vs. `SetVelocity(Vec2{x=1})` mean two different things but look the same on the wire.  So I see this wrapper pattern used, and I think it's a good one and an important one to enforce in the codebase for any "nullable" fields.

Ref: https://stackoverflow.com/a/50099927

Whether `PartialVec2` is actually a useful data type right now is a reasonable question -- I added it mostly to have this conversation. You can find a similar approach BD's [Spot API](https://github.com/boston-dynamics/spot-sdk/blob/master/protos/bosdyn/api/geometry.proto#L254).

python/farm_ng/tractor/sim/__main__.py
isherman: Will add.

isherman:
> For some reason the `yarn dev-server` command no longer works. Will try to understand what changed, but if you have an updated readme would be helpful.

Will add a README.

scripts/build-protos.sh
isherman: Cool, glad you like the approach. I don't have any experience with monolithic dev docker.

isherman: On the downside, CI time is more than it should be after this PR (~ 4mins). Will need to invest a few more cycles in caching these Docker images on CI but I don't think that will be difficult.

python/farm_ng/tractor/sim/models/waypoint.py
isherman: Yeah, another topic for discussion. Was hoping to think through this with you. 

In my last project we had serialization logic (to a. the network and b. the database) in a single model. And then business logic implemented as functions that operated on those models (rather than business logic in the model itself).

And it occasionally caused issues, I'm trying to remember exactly what they were. I think, for example, if you wanted to change something about the API, you ended up making changes to the model class, which often meant updating persistence logic too. 

While trying to understand how it could have been built better I would often think about [this HN thread](https://news.ycombinator.com/item?id=18189339):
> "...everyone always thinks it’s a good idea to tie the model object to the serializer object at the beginning of a project, and by the time you are ready to build a time machine to kill the person who made that mistake, it’s much too late to do anything but rewrite the whole thing"

And so I wanted to try to establish a pattern early on, in this PR, for separating the concerns of internal models from the API definitions. 

Reading it again now, I think the commenters are suggesting something different than what I implemented in this PR: pulling `toProto, fromProto` into their own functions outside of the model class. Which I can see an argument for.

I don't know, not trying to overengineer here. But I do think we should consider making internal APIs accept models, not generated proto classes. Different from the ROS approach. I suppose part of it depends on whether you're designing for high-performance pub/sub style use cases or more of an RPC use case. Also related to some of your early questions about serialization for logging.

python/farm_ng/tractor/sim/__main__.py
isherman: Done

app/frontend/test/api.int.test.ts
isherman: Cool! Going to leave this as is in this PR, if you're ok with it -- just needed some placeholder models to start with. Looking forward to starting to model the domain :)

python/farm_ng/tractor/sim/models/waypoint.py
isherman: Yes, that was the intention. Points to another issue with duplicating these models like this -- documentation.

python/farm_ng/tractor/sim/models/waypoint.py
isherman: Another pattern to throw into the mix is [this one](https://developers.google.com/protocol-buffers/docs/javatutorial):
> Protocol Buffers and O-O Design Protocol buffer classes are basically dumb data holders (like structs in C); they don't make good first class citizens in an object model. If you want to add richer behaviour to a generated class, the best way to do this is to wrap the generated protocol buffer class in an application-specific class. Wrapping protocol buffers is also a good idea if you don't have control over the design of the .proto file (if, say, you're reusing one from another project). In that case, you can use the wrapper class to craft an interface better suited to the unique environment of your application: hiding some data and methods, exposing convenience functions, etc. You should never add behaviour to the generated classes by inheriting from them. This will break internal mechanisms and is not good object-oriented practice anyway.

isherman: And one more [discussion](https://softwareengineering.stackexchange.com/questions/170760/protobuf-design-patterns) on this topic.

app/frontend/README.md
```diff
@@ -14,6 +14,6 @@ yarn
 
 ### Launch the web app
 ```
-yarn dev-server
+yarn dev-start
```
ethanrublee, line 17: +1

app/frontend/test/api.int.test.ts
ethanrublee: Got it sounds good.

python/farm_ng/tractor/sim/models/waypoint.py
ethanrublee: I vote for direct use of protobufs when possible as datatypes in remote APIs, and serialized forms, rather than translation to a very similar data model class defined in the language, and maybe even as the data model used in non-relational databases (if we go that route). 

Thoughts:
1. Maintaining model classes by hand in each of the respective languages (c++, python, javascript) leads to a lot of overhead for developers when they make changes in the protobuf, as it results in manually updating to/from proto definitions in each of the respective languages.
2.  Direct use of the generated protobuf types can be awkward, but it ends up being consistent and uses well documented idioms that translate pretty well across the languages, so its less cognitive load on developers, and they can copy patterns from one example to others, rather than learning a different api per each language.  Usually a simple library evolves in each respective language for common operations on the protobufs, which end up being compositions of stable basic objects such as poses, strings, and images, tensors.
3.  For math types and image types in particular, we often have to translate between some serialized binary representation to a datastructure, in a zero copy or simple memcopy fashion, for doing math, e.g. Eigen::Map, torch.Tensor, numpy.ndarray, gl buffer, and to and from proto definitions make sense here, or jpeg/mpeg encoding, etc.  I like the tensorflow Tensor datatypes as an example here.
4. For database tables, such as mongodb (not sure how this would work out for relational databases), Using a limited subset of protobuf and the json emitter works pretty well to serialize or serialize documents from the database.  But perhaps database models need separate treatment, to support complex queries and cross referencing.

scripts/build-protos.sh
ethanrublee: We could also have a repository for the docker files, and build them separately, as they may not change as frequently, just a thought for future.

PR #39, opened 2020-08-31
Extrinsic cal stubs

ethanrublee:
This adds apriltag, sophus, and ceres as dependencies, and a bit of a stub for the extrinsic calibration process, which just sends out a transform between tractor/base and the tracking_camera/front frame.


ethanrublee:
I may add apriltag detection to this branch.

cpp/farm_ng/extrinsic_calibrator.cpp
```diff
@@ -0,0 +1,92 @@
...
+    }
+    timer_.expires_from_now(boost::posix_time::millisec(200));
+    timer_.async_wait(
+        std::bind(&ExtrinsicCalibrator::on_timer, this, std::placeholders::_1));
+    bus_.Send(GenerateExtrinsicPoseEvent());
+  }
+
+  void on_event(const farm_ng_proto::tractor::v1::Event& event) {
+    if (event.data().type_url() !=
+        "type.googleapis.com/farm_ng_proto.tractor.v1.NamedSE3Pose") {
```
isherman, line 58: could be something like `"type.googleapis.com/" + farm_ng_proto::tractor::v1::NamedSE3Pose.descriptor().full_name()`

ethanrublee: thanks.  updated.

ethanrublee:
@isherman this currently runs detection and publishes it from the tracking_camera code.  Verified it runs on the tractor.

Suggest printing some april tags to test it out, hard coded for the tag36h11 group, you can find pngs here https://github.com/AprilRobotics/apriltag-imgs/tree/master/tag36h11

isherman:
> @isherman this currently runs detection and publishes it from the tracking_camera code. Verified it runs on the tractor.
> 
> Suggest printing some april tags to test it out, hard coded for the tag36h11 group, you can find pngs here https://github.com/AprilRobotics/apriltag-imgs/tree/master/tag36h11

sure, will make sure i can reproduce here before approving

cpp/farm_ng/tracking_camera.cpp
```diff
@@ -5,14 +5,29 @@
...
+#include <farm_ng/sophus_protobuf.h>
+
+#include <farm_ng_proto/tractor/v1/apriltag.pb.h>
 #include <farm_ng_proto/tractor/v1/geometry.pb.h>
 #include <farm_ng_proto/tractor/v1/tracking_camera.pb.h>
 #include <google/protobuf/util/time_util.h>
 
+#include <glog/logging.h>
+
+#include <chrono>
```
isherman, line 17: was getting `error: 'mutex' in namespace 'std' does not name a type` until I added `#include <mutex>`

would https://include-what-you-use.org/ have caught this?

third_party/readme.md
```diff
@@ -1,14 +1,51 @@
...
 ```
 
 Install udev rules:
-```
-sudo cp $FARM_NG_ROOT/third_party/librealsense/config/99-realsense-libusb.rules /etc/udev/rules.d/ 
+```bash
+sudo cp $FARM_NG_ROOT/third_party/librealsense/config/99-realsense-libusb.rules /etc/udev/rules.d/
 sudo udevadm control --reload-rules && udevadm trigger
 ```
+
```
isherman, line 15: I needed the following additional deps in `bootstrap.sh`:
```
libeigen3-dev \
libgoogle-glog-dev \
libopencv-dev \
```

protos/farm_ng_proto/tractor/v1/apriltag.proto
```diff
@@ -0,0 +1,47 @@
...
+    // the decision threshold. Higher numbers roughly indicate better
+    // decodes. This is a reasonable measure of detection accuracy
+    // only for very small tags-- not effective for larger tags (where
+    // we could have sampled anywhere within a bit cell and still
+    // gotten a good detection.)
+    float decision_margin = 4;
+
+    // The 3x3 homography matrix describing the projection from an
+    // "ideal" tag (with corners at (-1,1), (1,1), (1,-1), and (-1,
+    // -1)) to pixels in the image. This matrix will be freed by
```
isherman, line 33: Don't think the ownership note is relevant here

third_party/readme.md
```diff
@@ -1,14 +1,51 @@
...
 mkdir build-librealsense
 cd build-librealsense
 cmake -DBUILD_PYTHON_BINDINGS=true -DCMAKE_BUILD_TYPE=Release -DBUILD_EXAMPLES=false -DCMAKE_INSTALL_PREFIX=$FARM_NG_ROOT/env -DCMAKE_PREFIX_PATH=$FARM_NG_ROOT/env/ ../librealsense/
 ```
 
 Install udev rules:
-```
-sudo cp $FARM_NG_ROOT/third_party/librealsense/config/99-realsense-libusb.rules /etc/udev/rules.d/ 
+```bash
+sudo cp $FARM_NG_ROOT/third_party/librealsense/config/99-realsense-libusb.rules /etc/udev/rules.d/
```
isherman, line 12: FYI, I've been running a single T265 on the Jetson successfully without this installation step

cpp/farm_ng/tracking_camera.cpp
```diff
@@ -99,13 +114,18 @@ Event ToNamedPoseEvent(const rs2::pose_frame& rs_pose_frame) {
...
+  vodom_pose_t265.set_frame_a("tracking_camera/front");
   auto pose_data = rs_pose_frame.get_pose_data();
   SetVec3FromRs(vodom_pose_t265.mutable_a_pose_b()->mutable_position(),
                 pose_data.translation);
   SetQuatFromRs(vodom_pose_t265.mutable_a_pose_b()->mutable_rotation(),
                 pose_data.rotation);
+
+  Sophus::SE3d se3;
+  ProtoToSophus(vodom_pose_t265.a_pose_b(), &se3);
+  SophusToProto(se3.inverse(), vodom_pose_t265.mutable_a_pose_b());
```
isherman, line 127: what's the thinking around this change?

cpp/farm_ng/tracking_camera.cpp
```diff
@@ -114,19 +134,81 @@ Event ToNamedPoseEvent(const rs2::pose_frame& rs_pose_frame) {
...
+    tag_detector_->debug = false;
+    tag_detector_->refine_edges = true;
+  }
+
+  ~ApriltagDetector() {
+    apriltag_detector_destroy(tag_detector_);
+    tag36h11_destroy(tag_family_);
+  }
+
+  ApriltagDetections Detect(cv::Mat gray) {
```
isherman, line 157: better name than `gray`?

cpp/farm_ng/tracking_camera.cpp
```diff
@@ -114,19 +134,81 @@ Event ToNamedPoseEvent(const rs2::pose_frame& rs_pose_frame) {
...
+    image_u8_t im = {.width = gray.cols,
+                     .height = gray.rows,
+                     .stride = gray.cols,
+                     .buf = gray.data};
+
+    ApriltagDetections pb_out;
+    zarray_t* detections = apriltag_detector_detect(tag_detector_, &im);
+    // Draw detection outlines
+    for (int i = 0; i < zarray_size(detections); i++) {
+      ApriltagDetection* detection = pb_out.add_detections();
```
isherman, line 170: nit: would move this two lines down for easier top-to-bottom readability (get the detection, then populate the proto)

cpp/farm_ng/tracking_camera.cpp
```diff
@@ -114,19 +134,81 @@ Event ToNamedPoseEvent(const rs2::pose_frame& rs_pose_frame) {
...
+
+    // Make an image_u8_t header for the Mat data
+    image_u8_t im = {.width = gray.cols,
+                     .height = gray.rows,
+                     .stride = gray.cols,
+                     .buf = gray.data};
+
+    ApriltagDetections pb_out;
+    zarray_t* detections = apriltag_detector_detect(tag_detector_, &im);
+    // Draw detection outlines
```
isherman, line 168: not actually drawing anything here, right?

cpp/farm_ng/tracking_camera.cpp
```diff
@@ -188,10 +304,16 @@ class TrackingCameraClient {
   rs2::pipeline pipe_;
   std::mutex mtx_;
```
isherman, line 305: always appreciate a more descriptive name or, if not possible, comment to describe what's being protected

cpp/farm_ng/tracking_camera.cpp
```diff
@@ -163,11 +260,30 @@ class TrackingCameraClient {
...
     if (rs2::frameset fs = frame.as<rs2::frameset>()) {
       rs2::video_frame fisheye_frame = fs.get_fisheye_frame(0);
+      fisheye_frame.keep();
       cv::Mat frame_0 = frame_to_mat(fisheye_frame);
       std::lock_guard<std::mutex> lock(mtx_);
       count_ = (count_ + 1) % 3;
       if (count_ == 0) {
         writer_->write(frame_0);
+        if (!detection_in_progress_) {
+          detection_in_progress_ = true;
```
isherman, line 270: do we need to lock before changing this?

protos/farm_ng_proto/tractor/v1/geometry.proto
```diff
@@ -43,3 +43,9 @@ message NamedSE3Pose {
   string frame_a = 2;
   string frame_b = 3;
 }
+
+message Matrix {
```
isherman, line 47: not sure `geometry.proto` is the right home for this long term

isherman:
confirmed apriltag detection working (saw an apriltag event with the correct tag ID when running `farm-ng-ipc-logger`)
 
definitely felt some of the pain of non-hermetic development environment reviewing this. i think we should make some investments in this area soon, since we have deps spread across readmes, bootstrap.sh, all with side effects on the local system. the name `bootstrap.sh` is misleading, if it needs to be run everytime an additional system dependency is required.

a link to a CPP style guide might be useful as the number of contributors grows. i would say the naming style so far trends towards short names & abbreviations more than my personal preference, but i'm sure you have strong opinions on that topic.

also minor note, struggled a bit to find a workflow that would let me suggest a set of local changes on top of this PR. since the branch is in your personal forked repo. i can't push to it. and couldn't find an alternative blessed github workflow to get that diff somewhere you could just 'accept changes'.

cpp/farm_ng/tracking_camera.cpp
ethanrublee: I'm in favor of adding iwyu to our pre-commits, and/or as a build step in cmake.  Let's do that in a future PR.

third_party/readme.md
ethanrublee: Ok will add.

protos/farm_ng_proto/tractor/v1/apriltag.proto
ethanrublee: Thanks, sorry these comments are from apriltag library.

cpp/farm_ng/tracking_camera.cpp
ethanrublee: At the moment, the visualization code only supports displaying poses which have their reference frames in a specific order.  Ideally the client would be smart enough to traverse the tree, and be able to apply the rule that a_pose_b == b_pose_a.inverse().

So yeah this is a hack.

cpp/farm_ng/tracking_camera.cpp
ethanrublee: Need some image name conventions i guess.  gray is a pretty common name for arguments which expect a grayscale image.  Open to suggestions, suggest a follow up pass once we have a basic style guide.

cpp/farm_ng/tracking_camera.cpp
ethanrublee: This is covered by the scope of the lock above.

third_party/readme.md
ethanrublee: I think the debian install takes care of this for us.

ethanrublee: Removing the submodule and these lines from the readme.

ethanrublee:
Thinking of using this style guide:
https://google.github.io/styleguide/cppguide.html

Would be helpful to have some hooks in pre-commit to check clang-format, iwyu, and any style nits.  But let's save for a little later - suggest we make the build/deps handling better, and then incorporate these items.  For now, we can review for google style readability.

ethanrublee:
Ok, i made a pass at addressing review comments Ian. Also, I added you as a contributor to my fork, so if you would like to push some changes, you can add me as a remote with push access.

```
git remote add ethanrublee git@github.com:ethanrublee/tractor.git
```

isherman:
> Ok, i made a pass at addressing review comments Ian. Also, I added you as a contributor to my fork, so if you would like to push some changes, you can add me as a remote with push access.
> 
> ```
> git remote add ethanrublee git@github.com:ethanrublee/tractor.git
> ```

Cool, I think this is easier for now than an additional fork

cpp/farm_ng/tracking_camera.cpp
```diff
@@ -274,13 +286,15 @@ class TrackingCameraClient {
...
           io_service_.post([this, fisheye_frame, stamp] {
+            // note this function is called later, in main thread, via io_service_.run();
 	    cv::Mat frame_0 = frame_to_mat(fisheye_frame);
             Event event = farm_ng::MakeEvent("tracking_camera/front/apriltags",
                                              detector_.Detect(frame_0));
             *event.mutable_stamp() = stamp;
 
             event_bus_.Send(event);
-            std::lock_guard<std::mutex> lock(mtx_);
+	    // signal that we're done detecting, so can post another frame for detection.
```
isherman: some funky indentation in this file

PR #44, opened 2020-08-31
Cleanup

ethanrublee:


mikeferguson:
You'll want to update the README - it's linking missing files.

Also, where is all the rtkrover stuff going?

ethanrublee:
At the moment, we're focusing on camera based guidance, as we're running into cases where the RTK based navigation either isn't working well (near trees, or under greenhouses), and it requires a reliable network infrastructure to work properly.

I'll bring the RTK stuff back once we're able to test it regularly, but not sure when that will be.

ethanrublee:
@mikeferguson Thanks, point taken on the README.  We'll add a more clear guide shortly.

mikeferguson:
> @mikeferguson Thanks, point taken on the README

I'm notorious for leaving broken links in my READMEs, so it's something I always check

.gitmodules
```diff
@@ -7,3 +7,6 @@
 [submodule "third_party/protoc-gen-validate"]
 	path = third_party/protoc-gen-validate
 	url = https://github.com/envoyproxy/protoc-gen-validate.git
+[submodule "jetson/uhubctl"]
```
isherman, line 10: surprised to see this in this cleanup PR... i haven't needed it, i assume it's something specific to your setup

jetson/bringup.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash -ex
```
isherman, line 1: should this be inside `scripts`, or do you foresee `jetson` being an important top-level dir?

jetson/install.sh
```diff
@@ -9,5 +9,15 @@ while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symli
...
+prefix=/opt/farm_ng make -C $SERVICE_DIR/uhubctl
+prefix=/opt/farm_ng make -C $SERVICE_DIR/uhubctl install
+cp $SERVICE_DIR/*.sh /opt/farm_ng/systemd
 cp $SERVICE_DIR/*.service /etc/systemd/system/
+
 systemctl daemon-reload
+
+# start on boot always...
+systemctl enable tractor-bringup.service
+systemctl enable tractor-steering.service
```
isherman, line 22: slightly concerned about drift over time, for example if we want to remove a service. is there a way to, for example, put all the services inside a namespace/prefix, wipe that namespace at the beginning of each install, then add the services?

.gitmodules
isherman: nevermind, saw the motivation later in the PR

jetson/bringup.sh
ethanrublee: Not sure if jetson is the correct name, but basically this is the always on, systemd configuration for the embedded sides.  It could go under scripts, but would like to keep these scripts together as its a central view of the system configuration for the nano.  At the moment, its frankenstein as some of the scripts depend on the source code in /home/farmer/tractor --- so need to add an install python, cmake, and typescript.

jetson/install.sh
ethanrublee: I'm kind of new to the systemd side of things, and not sure how to namespace these.

We can remove all services named tractor*.service, and the support scripts are installed to /opt/farm_ng/ so this.  They may be a way to make this better, like through a debian package which can be uninstalled...

PR #52, opened 2020-09-02
Multicast and bringup fixes.

ethanrublee:
- Use a systemd.path unit to detect when the canbus and usb are ready
from tractor-bringup.service
- Enable multicast on loopback, in case the robot is not connected to
wifi.


isherman (approved):
would need to read more about system.d to give a proper review, but doesn't look very controversial

does `/tmp/tractor-ready.touch` ever need to be cleared?

ethanrublee:
/tmp/tractor-ready.touch gets removed at boot time.  Need more experience with systemd, so this is a work in progress that needs some mileage.

ethanrublee:
I just ran this around my half mile loop, and didn't have any emergency stop conditions.  So its an improvement....

PR #57, opened 2020-09-04
Compute distance traveled.

ethanrublee:


protos/farm_ng_proto/tractor/v1/tractor.proto
```diff
@@ -0,0 +1,7 @@
+syntax = "proto3";
+
+package farm_ng_proto.tractor.v1;
+
```
isherman, line 4: please add `option go_package = "github.com/farm_ng/genproto"`

see https://github.com/farm-ng/tractor/issues/58

python/farm_ng/kinematics.py
```diff
@@ -20,8 +20,11 @@ def unicycle_to_wheel_velocity(self, v: float, w: float) -> Tuple[float, float]:
...
     def evolve_world_pose_tractor(
         self, world_pose_tractor: SE3, v_left: float,
-        v_right: float, delta_t: float,
+        v_right: float, delta_t: float, delta_pose: bool = False,
     ) -> SE3:
         v, w = self.wheel_velocity_to_unicycle(v_left, v_right)
         tractor_pose_t = SE3.exp([v*delta_t, 0, 0, 0, 0, w*delta_t])
-        return world_pose_tractor.dot(tractor_pose_t)
+        world_pose_tractor = world_pose_tractor.dot(tractor_pose_t)
+        if delta_pose:
```
isherman, line 28: the overloaded return type is confusing -- can we always return both?

protos/farm_ng_proto/tractor/v1/tractor.proto
```diff
@@ -0,0 +1,7 @@
...
+message TractorState {
+  double abs_distance_traveled = 1;
```
isherman, line 6: maybe add a comment this is just since the last software restart (typically since the last power cycle)

protos/farm_ng_proto/tractor/v1/tractor.proto
ethanrublee: Added, this but fails at lint time:
```
Files in package "farm_ng_proto.tractor.v1" have both values "github.com/farm_ng/genproto" and no value for option "go_package" and all values must be equal.
```

I think this change needs to be made after go is added to the master branch.

protos/farm_ng_proto/tractor/v1/tractor.proto
ethanrublee: Added comment, thanks for the note, was thinking about the reset caveat as well.  Not sure best way or if its necessary to implement a true odometer.

python/farm_ng/kinematics.py
```diff
@@ -18,10 +18,14 @@ def unicycle_to_wheel_velocity(self, v: float, w: float) -> Tuple[float, float]:
         v_left = (2*v - w*self.wheel_base_line)/(2*self.wheel_radius)
         return v_left, v_right
 
+    def compute_tractor_pose_delta(self, v_left: float, v_right: float, delta_t: float) -> SE3:
```
ethanrublee, line 21: yes this is more clear, thanks!

ethanrublee:
Think this is ready for another review.

python/farm_ng/kinematics.py
isherman: oh, interesting, I was thinking 
```
# callsite wants both evolved pose and the computed delta
self.odom_pose_tractor, tractor_pose_delta = kinematics.compute_tractor_pose_delta(self._left_vel, self._right_vel, dt)

# callsite only cares about the evolved pose
self.odom_pose_tractor, _ = kinematics.compute_tractor_pose_delta(self._left_vel, self._right_vel, dt)
```

then you avoid leaking / duplicating the `self.odom_pose_tractor = self.odom_pose_tractor.dot(tractor_pose_delta)` math into `tractor.py`

isherman (approved):

python/farm_ng/kinematics.py
ethanrublee: I felt like the math was clear enough that it wasn't worth the complexity of hiding it. But we can revisit

PR #67, opened 2020-09-06
Add simple python wifi manager

isherman:
A small tool to enable switching `wlan0` between acting as a hotspot or connecting to an existing wifi network.

Could be done solely with bash scripts, but with a Python interface we get closer to exposing this functionality to end users, and supporting features like a 'captive portal', dns, etc.

https://github.com/balena-io/wifi-connect, and its Python implementation, https://github.com/OpenAgricultureFoundation/python-wifi-connect, are two inspirations.

The main limitation of this implementation at the moment is it expects you to have connected to a wifi network at least once so that network manager has it saved as a connection.


jetson/readme.md
```diff
@@ -37,4 +35,26 @@ journalctl -f -u tractor-bringup
...
 ```
 
-Or ``tail -f /var/log/syslog`` ...
+Or `tail -f /var/log/syslog` ...
+
+## Wifi Configuration
+
+### Enable the tractor as an access point
+
+- Run `python -m farm_ng.wifi ap`.
```
ethanrublee, line 44: Will this persist across reboots?

ethanrublee (commented):
Cool, will try it out.

jetson/readme.md
ethanrublee: One note, i had to do this as root.

ethanrublee: Also, does not persist across reboots.

ethanrublee (commented):
image

Streaming video + data over wifi hotspot.

In my garage, with wall and shelves between seem to be limited to about 1 mbit/second.

Open notes:
- How to make this persist across reboots? Add it to bringup.sh?

For multi-cast to work across this link (don't know if we care to support this), i had to re-add the multi-cast route after starting the AP.

python/farm_ng/wifi.py
```diff
@@ -0,0 +1,157 @@
...
+def print_connections():
+    connections = NetworkManager.Settings.ListConnections()
+    for c in connections:
+        print(c.GetSettings()['connection']['id'])
+
+
+if __name__ == '__main__':
+    logging.basicConfig(stream=sys.stdout, level=logging.INFO)
+
+    if len(sys.argv) < 2:
```
ethanrublee, line 136: Suggest using argparse.

python/farm_ng/wifi.py
```diff
@@ -0,0 +1,157 @@
...
+            'mode': 'ap',
+            'security': '801-22-wireless-security',
+            'ssid': ap_SSID,
+        },
+        '802-11-wireless-security': {
+            'key-mgmt': 'wpa-psk',
+            'psk': password,
+        },
+        'connection': {
+            'autoconnect': False,
```
ethanrublee, line 33: I have two wifi connections (home, and hotspot) i want to toggle "persistently" between.  Maybe toggling connection: autoconnect on both would work well?  for example in hotspot mode, autoconnect for hotspot is true, autoconnect for home wifi is false.

python/farm_ng/wifi.py
isherman: done

python/farm_ng/wifi.py
isherman: yup, this was an oversight. managed persistently with 'autoconnect' now.

isherman:
> For multi-cast to work across this link (don't know if we care to support this), i had to re-add the multi-cast route after starting the AP.

Can you send me the contents of your `/etc/NetworkManager/system-connections/<farm_ng-hostname>` file (after you added the multicast route)? I should be able to add the multicast route to the access point config.

jetson/readme.md
isherman: > Will this persist across reboots?

Ack, sorry this was an oversight in the initial PR. Tested and working on my Jetson now.

> One note, i had to do this as root.

Interesting, I haven't had to. Would be good to debug that together real quick.

isherman:
> In my garage, with wall and shelves between seem to be limited to about 1 mbit/second.

That's not great for video -- a note I have here is 360p at 30FPS will saturate that. Do you know the model of the wireless NIC we have in there?

python/farm_ng/wifi.py
```diff
@@ -0,0 +1,184 @@
...
+
+def print_connections():
+    connections = NetworkManager.Settings.ListConnections()
+    for c in connections:
+        print(c.GetSettings()['connection']['id'])
+
+
+class WifiCLI:
+    def __init__(self):
+        parser = argparse.ArgumentParser(
```
ethanrublee, line 145: This impl is fine, but for reference there is support for sub commands in argparse:
https://docs.python.org/3/library/argparse.html#sub-commands

jetson/readme.md
ethanrublee: I call it as root using:
```
farmer@tractor0003:~/tractor$ sudo ./env.sh python -m farm_ng.wifi ap
```

I think its Ok for now, we'll figure out groups and permissions later.

ethanrublee (commented):
The routes don't show up in my network manager config (haven't figured out/researched how to do that but it sounds like a good way to go).

I manually add the route with this command:
jetson/bringup.sh:route add -net 224.0.0.0 netmask 240.0.0.0 dev wlan0

ethanrublee (approved):

python/farm_ng/wifi.py
```diff
@@ -0,0 +1,184 @@
...
+    # Get the wireless network interface
+    device = next((d for d in NetworkManager.NetworkManager.GetDevices() if d.Interface == wifi_interface), None)
+    if not device:
+        raise Exception(f'No {wifi_interface} device found')
+
+    # Enable autoconnect for selected connection only
+    wifi_connections = [c for c in NetworkManager.Settings.ListConnections() if c.GetSettings()['connection']['type'] == '802-11-wireless']
+    for c in wifi_connections:
+        c_settings = c.GetSettings()
+        c_settings['connection']['autoconnect'] = (c_settings['connection']['id'] == connection_id)
```
ethanrublee, line 83: works well on my machine, thanks! tried reboot persistence for both infra/ap.

python/farm_ng/wifi.py
isherman: ack, didn't see that. thanks for the heads-up

PR #73, opened 2020-09-08
Clamp dt as its sometimes negative, work around issue #53

ethanrublee:


ethanrublee:
Ready for review I think.

ethanrublee:
Runs on tractor.

isherman (approved):

PR #75, opened 2020-09-08
Tractor config proto 68

ethanrublee:
Adds stub of tractor config, and a sample program to print json (but no persistence on disk).

Also, this fixes a units inconsistency between hub motor and the kinematic model, now that they both share the wheel_radius value.

### DIFF
--- a/protos/farm_ng_proto/tractor/v1/tractor.proto
+++ b/protos/farm_ng_proto/tractor/v1/tractor.proto
@@ -1,7 +1,6 @@
 syntax = "proto3";
 
 import "google/protobuf/wrappers.proto";
-import "farm_ng_proto/tractor/v1/geometry.proto";
 
 package farm_ng_proto.tractor.v1;
 option go_package = "github.com/farm_ng/genproto";
@@ -23,6 +22,4 @@ message TractorConfig {
   // Hub motor polls
   google.protobuf.Int32Value hub_motor_poll_pairs = 4;
 
-  // Extrinsic relationship of sensors and motors.
-  PoseTree extrinsics = 5;
 }
--- a/protos/farm_ng_proto/tractor/v1/tractor.proto
+++ b/protos/farm_ng_proto/tractor/v1/tractor.proto
@@ -1,5 +1,8 @@
 syntax = "proto3";
 
+import "google/protobuf/wrappers.proto";
+import "farm_ng_proto/tractor/v1/geometry.proto";
+
 package farm_ng_proto.tractor.v1;
 option go_package = "github.com/farm_ng/genproto";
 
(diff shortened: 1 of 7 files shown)


ethanrublee:
FYI this validated on the robot to my best ability.

protos/farm_ng_proto/tractor/v1/tractor.proto
```diff
@@ -9,3 +12,17 @@ message TractorState {
...
+  google.protobuf.DoubleValue wheel_baseline = 1;
+  // Radius of wheel, in meters.
+  google.protobuf.DoubleValue wheel_radius = 2;
+  // Hub motor gear ratio
+  google.protobuf.DoubleValue hub_motor_gear_ratio = 3;
+  // Hub motor polls
+  google.protobuf.Int32Value hub_motor_poll_pairs = 4;
+
+  // Extrinsic relationship of sensors and motors.
+  PoseTree extrinsics = 5;
```
isherman, line 27: I worry a bit about losing the type-safety of the names of the nodes in the tree. We'll be relying on string matching. Something more explicit that represents the tractor's expected calibration frames might be better. Thoughts?

ethanrublee: I agree - need some way to enforce the poses.  However, having a more generic representation makes the library/debugging tools a bit more flexible/decoupled from the application.  In general its hard to enforce invariants in protobuf.

A config check could enforce expected frames/connections in the datastructure.

I think likely this field is consumed and updated by a generic calibration routine.

We could use introspection in code to enable parsing of field names to contruct a pose tree datastructure.

```protobuf
SE3Pose tractor_base_pose_tracking_camera_front = 1;
SE3Pose tractor_base_pose_tracking_camera_rear = 2;
```

But I think a simple function for ensuring a pose tree has the expected frames is sufficient.  And then we can more easily leverage library code that performs on the tree data-structure.

ethanrublee: This field isn't used in this pull request, so could remove it and consider this later when we're using it during calibration/visualization.

isherman: I see the point. Though i think you could argue the library code layer should provide PoseTree<->TractorPoseTree conversion, for libraries that consume generic PoseTrees. Then we have the guarantees, for example, that any tractor config file that gets saved to disk has all the necessary reference frames (without enforcing it by convention in library code). I guess I think of it kind of like modeling a database, where there are benefits to referential integrity in the DB (the SQL approach) vs. enforcing it all in application code (the schemaless NoSQL approach).

isherman (commented):
This looks ready to merge to me, but will let @mmcgowen2015 give approval after testing on hardware

protos/farm_ng_proto/tractor/v1/tractor.proto
ethanrublee: I'm open to this idea. I think for now, i'll remove extrinsics and then we can iterate on a type safe convention with the first usecase for the PoseTree (when we land some camera/base calibration).

mabitter (approved):
Sweet.

PR #94, opened 2020-09-14
Simplify switch and install scripts

isherman:
## Proposed change to wiki

```bash
# Preconditions:
#   - Ensure the tractor is on a wifi network with internet access
#   - Ensure there are no local changes
# Examples: 
#  ./scripts/switch.sh 999
#  ./scripts/switch.sh master
./scripts/switch.sh <PR# or master>

# Update system dependencies, update startup scripts, build the latest code
make

# Make any necessary configuration changes
# Examples:
#   ./env.sh python -m farm_ng.wifi ap

# Restart the tractor
sudo shutdown -r now
```

## Test Plan

- [x] Dirty working tree refuses to run
- [x] New system dependency in `bootstrap-apt.sh`
- [x] Added systemd script in `jetson/`
- [x] Removed systemd script in `jetson/`
- [x] Added dependency in `third_party`
- [x] Change in cpp code
- [x] Change in go code
- [x] Change in frontend code
- [x] Switches successfully to `PR#` 
- [x] Switches successfully to `master`


isherman:
Would like to add this `make` step to the Github CI in a follow-up PR (though it will slow down CI... thoughts?)

isherman:
> Would like to add this `make` step to the Github CI in a follow-up PR (though it will slow down CI... thoughts?)

We can discuss in https://github.com/farm-ng/tractor/pull/96

Makefile
```diff
@@ -3,20 +3,34 @@ ifdef TEST_FILTER
...
+	make cpp
+	make webservices
+	make systemd
+	echo "\nSuccessful full system build.\n"
+
+bootstrap:
+	./bootstrap.sh
+
+cpp:
+	make protos
```
ethanrublee, line 18: cpp doesn't depend on this step.

bootstrap-apt.sh
```diff
@@ -35,4 +35,23 @@ apt-get install -y \
      protobuf-compiler \
      python3-pip
 
+# Virtualenv
 pip3 install virtualenv
+
+# Go
+wget https://golang.org/dl/go1.15.1.linux-arm64.tar.gz -P /tmp/
```
ethanrublee, line 42: would be nice if this checked if the version matched before download.

jetson/install.sh
```diff
@@ -9,23 +9,37 @@ while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symli
 done
 export SERVICE_DIR=$( cd "$( dirname "${SOURCE}" )" >/dev/null 2>&1 && pwd )
 
+# install uhubctl
 mkdir -p /opt/farm_ng/systemd
 prefix=/opt/farm_ng make -C $SERVICE_DIR/uhubctl
 prefix=/opt/farm_ng make -C $SERVICE_DIR/uhubctl install
+
+# clean
+systemctl list-unit-files | grep "tractor" | awk '{print $1}' | xargs --no-run-if-empty -n1 sudo systemctl disable
```
ethanrublee, line 18: nice way to disable.

ethanrublee (approved):

bootstrap-apt.sh
isherman: good idea. going to revisit this script to improve performance once this approach is confirmed working for someone besides me... I think we can save about 1min off the build.

Makefile
isherman: ok, will remove in follow-up

PR #96, opened 2020-09-14
Add full build to Github CI

isherman:


isherman:
Hm, well we're at 15mins before we've even finished building `third_party`. A good use case for the Docker build cache. Let's discuss whether this addition to CI is useful at this point.

ethanrublee:
I don't think this is quite useful yet.

Would like to make the build better (third_party etc), and get some unit tests in for c++ before we do this.

isherman:
OK, closing.

PR #97, opened 2020-09-14
Add collect-logs.sh

isherman:
Any other logs you think would be useful for debugging?

Still thinking about how to make sharing/uploading somewhere low-friction


isherman:
Looked at some upload/share options. Curious if you have any creative thoughts. Some options:
- Public or semi-public service like pastebin.com
- Upload to Google Drive (requires some GCP configuration, presumably under the farm-ng GSuite account, and us to put an API key on the tractors)
- Upload to cloud storage (AWS, Google, etc.)

scripts/collect-logs.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+LOGDIR=`mktemp -d -t farm-ng-log-$(date +"%s")-XXX`
+
+systemctl list-unit-files | grep "tractor" | awk '{print $1}' | while IFS= read -r unit; do
+  logfile=$LOGDIR/$unit.log
+  echo "Writing to $logfile"
+  journalctl -b --output=short-iso-precise --no-pager -u $unit | head -1000 > $logfile
```
ethanrublee, line 8: For now we may want the entire log file, rather than just the last 1000 lines.

ethanrublee (changes requested):
Suggest bumping this to the whole available log, rather than the last 1000 lines.

scripts/collect-logs.sh
isherman: This is the first 1000 lines after boot. 

Unfortunately the full log since boot was, on my machine, 8 megabytes (all text from that 5s status loop). I guess it should compress to almost nothing (didn't look at the compressed size). But could be a bit heavy to view client-side. 

Will confirm it compresses well and make that change.

isherman: looks like a difference in the final .zip file between 50K and 1MB

ethanrublee (approved):

ethanrublee:
Do these attach well to issues? I guess they can potentially contain PII (gps coordinates), would be cool if we somehow have anonymous logs that are safe for posting on a public forum.

ethanrublee:
[log_test.zip](https://github.com/farm-ng/tractor/files/5222128/log_test.zip)


Looks like github allows for upload of zip files to tickets.

isherman:
> [log_test.zip](https://github.com/farm-ng/tractor/files/5222128/log_test.zip)
> 
> Looks like github allows for upload of zip files to tickets.

oh that's an interesting idea, will look at that

PR #83, opened 2020-09-10
Add skeleton for calibration library

ethanrublee:
- gtest support

PR #100, opened 2020-09-16
Define image and resource protos

isherman:


protos/farm_ng_proto/tractor/v1/image.proto
```diff
@@ -0,0 +1,45 @@
...
+option go_package = "github.com/farm_ng/genproto";
+
+message Image {
+  // The associated image resource
+  Resource resource = 1;
+
+  // General metadata
+  ImageMetadata metadata = 2;
+
+  // CalibrationMetadata (objects, apriltag detections, reprojections, rmse),
```
ethanrublee, line 17: It seems to me that the CalibrationMetadata doesn't belong here, as Image is likely to be used in a few applications, so CalibrationImage might contain Image.

protos/farm_ng_proto/tractor/v1/image.proto
```diff
@@ -0,0 +1,45 @@
...
+
+// Metadata common to a full image set
+message ImageSetMetadata {
+  // (Optional) The time the image set was assembled.
+  google.protobuf.Timestamp timestamp = 1;
+
+  // Camera intrinsics, etc.
+}
+
+message ImageSet {
```
ethanrublee, line 42: From the streaming perspective, my gut is telling me that ImageSet as a container of images will be lead to batching up of data before its displayed or processed. I have in mind an alternative where ImageSet is basically a unique id, with some meta data (camera intrinsics, what its for), and can be attached to each Image.


For example:

```protobuf
message ImageSetMetadata {
  int64 uuid = 1;
  string name = 2;
};
message CalibrationFrame {
  Image image = 1;
  SetMetadata set = 2;
  repeated ApriltagDetections = 3;
};
```

Then the camera service begins capturing a new set base on some incoming Event message (possibly containing the ImageSetMetaData with uuid, and some human readable name), and attaches this metadata to each CalibrationFrame.

isherman: That makes sense, but how do we support the workflow we described before:

1. Enable capture mode on the camera, results in an `ImageSet` saved to disk.
2. Via the frontend, load the `ImageSet` and build a `TagLibrary`.
3. Save the `TagLibrary` to on-robot disk.

What is the `ImageSet` that's saved to disk -- a log of `Image`s with the same `ImageSet` id? 
And what's the data type the webserver replies with for the frontend to load (needs to be wrapped up in a single RPC response)?

protos/farm_ng_proto/tractor/v1/image.proto
isherman: will do

protos/farm_ng_proto/tractor/v1/image.proto
ethanrublee: I think if this is an ipc-log, that would be convenient, as it will also have motor messages and odometry messages, which will be required for other calibration tasks.

Here's a timing diagram for what I had in mind.

calibration application
https://docs.google.com/drawings/d/1OIdN89rOXKJrJz-xFcfWQwTQq48-UHkJjjdaC57veis

ethanrublee:
@isherman Ok, this is up to date with the latest on #83

protos/farm_ng_proto/tractor/v1/apriltag.proto
```diff
@@ -1,50 +1,75 @@
...
+
+  // Size of the tag, outside corner to corner on one side, in meters.
+  double tag_size = 9;
 }
 
 message ApriltagDetections {
-	repeated ApriltagDetection detections=1;
+  repeated ApriltagDetection detections = 1;
+  Image image = 2;
+  double rmse = 3;
```
isherman, line 61: this is only used by calibration? if so, does this belong here (by your argument about not attaching CalibrationMetadata to an Image).

isherman, line 61: on a related note, if this does stay, I think there's real value in using wrapper types to express that this is an explicitly "nullable" field (and doesn't just happen to be missing due to version skew).

protos/farm_ng_proto/tractor/v1/apriltag.proto
```diff
@@ -1,50 +1,75 @@
...
+}
+
+message ApriltagRig {
+  message Entry {
```
isherman, line 65: "Entry" doesn't feel quite right as a name to me. Are 'rig' and 'entry' standard names in the literature? "Rig" sounds like a physical object in the world, "Entry" sounds like a record in a database.

protos/farm_ng_proto/tractor/v1/apriltag.proto
```diff
@@ -1,50 +1,75 @@
...
+    int32 id = 1;
```
isherman, line 66: i notice there's not much use of `uint`s in the new protobuf definitions (for ids, width, height). is this by design (like allowing -1 as a null value) or coincidence? i don't have strong feelings either way, just want to make sure i know the convention if we're adopting one implicitly.

isherman: my slight preference would probably be for `Google.Protobuf.WellKnownTypes.UInt32Value` if the intention is "an id we use for bookkeeping that can be null"

protos/farm_ng_proto/tractor/v1/apriltag.proto
```diff
@@ -1,50 +1,75 @@
...
+    string frame_name = 2;
+    double tag_size = 3;
```
isherman, line 68: pls document units (either here or globally for the project)

isherman: a short docstring for all these would be helpful actually (like what coordinate frame is `tag_points` in?)

protos/farm_ng_proto/tractor/v1/calibrator.proto
```diff
@@ -0,0 +1,57 @@
...
+  SOLVER_STATUS_CONVERGED = 2;
+  SOLVER_STATUS_FAILED = 3;
+}
+
+message ApriltagRigTagStats {
+  int32 tag_id = 1;
+  // how many frames was this tag observed in
+  int32 n_frames = 2;
+  double tag_rig_rmse = 3;
+  map<int32, double> per_image_rmse = 4;
```
isherman, line 32: what are the keys in this map? 
is `image` here different from `frame` as used two lines above?

protos/farm_ng_proto/tractor/v1/calibrator.proto
```diff
@@ -0,0 +1,57 @@
...
+}
+
+message MonocularApriltagRigModel {
+  ApriltagRig rig = 1;
+  SolverStatus solver_status = 2;
+  double rmse = 3;
+  repeated ApriltagRigTagStats tag_stats = 4;
+  string camera_frame_name = 5;
+  repeated ApriltagDetections detections = 6;
+  PoseTree camera_poses_rig = 7;
```
isherman, line 42: this name confuses me

protos/farm_ng_proto/tractor/v1/apriltag.proto
ethanrublee: Good point, let me review in the calibration code how i used this field and see.

It can be computed directly from the apriltag pose estimation routine and its valid to put here in my option.

protos/farm_ng_proto/tractor/v1/calibrator.proto
```diff
@@ -0,0 +1,57 @@
...
+  repeated Image reprojection_images = 8;
+}
+
+message CalibratorStatus {
+  message ApriltagRigProgress {
+    int32 num_frames = 1;
+    MonocularApriltagRigModel rig_model = 2;
+    // The rig model saved to disk.
```
isherman, line 50: cool. is this a new category of persistent blobs? what's in the file, binary protobuf?

protos/farm_ng_proto/tractor/v1/apriltag.proto
ethanrublee: These really can't be null, and its very onerous to use the value types directly in the detection/calibration code.

ints are used at the API level of apriltag and opencv width/height, so remaining consistent with signs prevents casting.

protos/farm_ng_proto/tractor/v1/camera_model.proto
```diff
@@ -0,0 +1,36 @@
...
+package farm_ng_proto.tractor.v1;
+option go_package = "github.com/farm_ng/genproto";
+
+message CameraModel {
+  // Image width in pixels.
+  int32 image_width = 1;
+  // Image height in pixels.
+  int32 image_height = 2;
+
+  // Image projection center x cooridinate in pixels, typically close to
```
isherman, line 12: typo: "coordinate" (and below)

protos/farm_ng_proto/tractor/v1/apriltag.proto
ethanrublee: Rig is somewhat common language for rigid collections of objects in pose estimation.  I'm not sure what to call the objects in the rig.  Rig::Entry seemed OK, but if you have a suggestion happy to refactor.

protos/farm_ng_proto/tractor/v1/camera_model.proto
```diff
@@ -0,0 +1,36 @@
...
+
+  enum DistortionModel {
+    DISTORTION_MODEL_UNSPECIFIED = 0;
+    // Intel Realsense T265 - Four parameter Kannala Brandt distortion model
+    DISTORTION_MODEL_KANNALA_BRANDT4 = 1;
+  };
+
+  // Specifies the distortion model.
+  DistortionModel distortion_model = 7;
+  // Distorion model coeffecients, typically has 5 values, some may be 0.0.
```
isherman, line 31: typo: "Distortion", "coefficients"

protos/farm_ng_proto/tractor/v1/io.proto
```diff
@@ -7,16 +7,50 @@ package farm_ng_proto.tractor.v1;
...
+      MODE_UNSPECIFIED = 0;
+      MODE_ALL_MESSAGES = 1;
+    }
+    // log name, location on disk is determined by this.
+    string name = 1;
+    Mode mode = 2;
+  }
+
+  oneof command {
+    bool record_stop = 2;
```
isherman, line 36: maybe `google.protobuf.Empty`?

protos/farm_ng_proto/tractor/v1/io.proto
```diff
@@ -7,16 +7,50 @@ package farm_ng_proto.tractor.v1;
...
+
+message LoggingStatus {
+  message Recording {
+    // Log name
+    string name = 1;
+    // Where on disk the log is located
+    string path = 2;
+    // Number of messages recorded
+    int64 n_messages = 3;
+    // Timestamp of the when the log started recording
```
isherman, line 49: typo: "the when the"

protos/farm_ng_proto/tractor/v1/calibrator.proto
ethanrublee: camera_poses_rig is a transform naming convention, where this means a collection of "poses", which describing the transform camera_pose_rig, which maps points in the rig frame to points in the camera frame (point_camera = camera_pose_rig*point_rig). need a global style guide on naming for poses, as this is quite important when writing/understanding what's going on in the math functions.

protos/farm_ng_proto/tractor/v1/io.proto
isherman: (and below, for `stopped`)

protos/farm_ng_proto/tractor/v1/apriltag.proto
ethanrublee: I should change this to "points_tag" to be consistent with pose algebra naming conventions..., these are points in the tag frame.

ethanrublee: And globally, units are meters (for points, and size).  Can document though.

protos/farm_ng_proto/tractor/v1/calibrator.proto
ethanrublee: these are the frame index in the apriltagdetections detections list in MonocularApriltagRigModel structure.

protos/farm_ng_proto/tractor/v1/calibrator.proto
ethanrublee: Its saved as a text/json document at the moment.  Perhaps the Resource should get the protobuf typeurl?

protos/farm_ng_proto/tractor/v1/calibrator.proto
```diff
@@ -0,0 +1,57 @@
...
+  string camera_frame_name = 5;
+  repeated ApriltagDetections detections = 6;
+  PoseTree camera_poses_rig = 7;
+  repeated Image reprojection_images = 8;
+}
+
+message CalibratorStatus {
+  message ApriltagRigProgress {
+    int32 num_frames = 1;
+    MonocularApriltagRigModel rig_model = 2;
```
ethanrublee, line 49: Might want to use only the resource as the rig_model could be large and not work well with streaming over the eventbus if we have datagram size limits.

protos/farm_ng_proto/tractor/v1/geometry.proto
```diff
@@ -45,8 +45,12 @@ message NamedSE3Pose {
   string frame_b = 3;
 }
 
+message PoseTree {
```
ethanrublee, line 48: I think i'll remove this type for now, and just use repeated NameSE3Pose.

protos/farm_ng_proto/tractor/v1/io.proto
ethanrublee: Ok.  Or i could define RecordStop message so if we need additional args to stop, version skew doesn't effect us.

protos/farm_ng_proto/tractor/v1/apriltag.proto
ethanrublee: I'll remove, as its not computed or used in the calibration code.

protos/farm_ng_proto/tractor/v1/calibrator.proto
isherman: that's a good question. if we need to encode the type i'd vote something like `application/json; type=<typeurl>` -- the semicolon convention is used by the HTTP Content-Type header

- https://stackoverflow.com/questions/30505408/what-is-the-correct-protobuf-content-type
- https://stackoverflow.com/questions/35761248/which-separator-should-be-used-in-the-content-type-header-for-a-multipart-data-r

protos/farm_ng_proto/tractor/v1/io.proto
isherman: an empty RecordStop message sounds reasonable. i like that better than the boolean

protos/farm_ng_proto/tractor/v1/calibrator.proto
isherman: got it

protos/farm_ng_proto/tractor/v1/apriltag.proto
isherman: ok

protos/farm_ng_proto/tractor/v1/apriltag.proto
isherman: Rig makes sense. one suggestion i like a bit better than `Entry` would be `Node`. i don't feel that strongly about it, just trying to put myself in the shoes of a developer integrating against the calibration library.

protos/farm_ng_proto/tractor/v1/calibrator.proto
isherman: (i guess they're called 'parameters') https://tools.ietf.org/html/rfc7231#section-3.1.1.1

ethanrublee:
seems like this can be closed as #102 was merged? I'll address the feedback in #83

protos/farm_ng_proto/tractor/v1/geometry.proto
ethanrublee: done.

ethanrublee:
See
https://github.com/farm-ng/tractor/pull/83/commits/9e1658b728842fe4dbcdd62bfcf3ea243f1ce186

protos/farm_ng_proto/tractor/v1/apriltag.proto
ethanrublee: Changed to Node.

protos/farm_ng_proto/tractor/v1/apriltag.proto
ethanrublee: Added start of design/conventions.md in #83 

Also added per protobuf comments here.

protos/farm_ng_proto/tractor/v1/calibrator.proto
ethanrublee: Done, thanks for the suggestion, added convenience function in c++ to write protobuf files consistently out as json with the mime type. Can extend later for binary.

Issue #88, opened 2020-09-11
Add generic visualizer


isherman:
Some thoughts about what a generic visualizer for Eventbus events could look like. 

This approach supports visualization of individual `Event`s, and vectors of events `{..., En-1, En, En+1, ...}` by `Stream`, where a `Stream` is defined by a {`DataType`, `Tag`}. `Tag` would, to start, be equivalent to the event source (`camera`, `steering`, etc.), but in the future it could be a field populated by the application.

The data model is inspired by tools like OpenCV or matplotlib's `imshow`, and [Tensorboard](https://www.tensorflow.org/tensorboard).

This is contrast to an approach like [Rviz](http://wiki.ros.org/rviz), which synthesizes multiple data streams into a single 3D scene. The analogue of the approach described here in the ROS ecosystem might be "rqt_plot for all types". An Rviz-like approach relies on a stateful, streaming protocol. Such a tool might be necessary in the future, and could be designed in parallel.

![farm-ng visualizer wireframes (1)](https://user-images.githubusercontent.com/601482/92963059-b755a780-f426-11ea-8b18-7fd29b29c59c.png)

1. The URL encodes the view configuration, allowing for links from other parts of the web app.

2. User can choose what data source should populate the backing buffer: live data (data will expire out of the back of the buffer), paused (data will not be added to the front of the buffer or expired out of the back), or log (the buffer is fully replaced by a contents of a log file).

3. Buffer visualizer shows buffer start time, event vector start time, event cursor time, event vector end time, buffer end time (infinity in live mode), sampling rate, and buffer size.

4. The data _type_ is at the top level of the visualizer hierarchy (again, inspired by Tensorboard). This allows us to set visualization options for all streams of this type together. User can input a filter on which `Tag`s should be displayed. A user can choose between various visualization types for a given data type (e.g. {plot, histogram} for scalars). "Inferred" is always available, which will walk down from the root of the `Event` structure until fields of a type with a registered visualization are found, and render those fields individually.  
The plus button on the right will create a new top-level panel, which allows the user to look at two (or more) types of data at once.

5. Within the top-level visualization panel we see each `Stream` rendered separately.

6. Within a `Stream`, we see either a list of panels or a single panel, depending on the visualization configuration (e.g. `overlay` vs. `grid` for images). The visualization implementation operates on a vector of Events, so it can choose to summarize / collapse them, or render them individually. These panels display timestamps and any other event-level metadata by default.

[...]

PR #102, opened 2020-09-18
Add v1 of oscilloscope

isherman:


isherman:
This is a first pass at the design proposed in https://github.com/farm-ng/tractor/issues/88.

Screen Shot 2020-09-22 at 12 36 34 AM

It's ready for some real world use to prioritize next steps.

I'm pretty happy with the visualizer registry, and the ease of adding new visualizers. Performance w/ log files in the 100s of megabytes feels snappy. Streaming at 30FPS is working, for high-frequency numeric data at least. Responsive design is on the right track, but will need refinement based on feedback. 3D visualization is still an open question.

Some known issues:
- No support for streaming images yet (i.e. remote resource archive).
- In "Overlay" and "Grid" views, newest data is on the last page rather than the first.
- Plots don't resize in response to panel subdivision. Workaround is to resize the window. Fix is the [Resize Observer API](https://developer.mozilla.org/en-US/docs/Web/API/Resize_Observer_API).

Performance will be an ongoing concern.

- [ ] Merge latest proto definitions before merging

app/frontend/src/components/scope/visualizers/ApriltagDetectionsVisualizer.tsx
```diff
@@ -0,0 +1,119 @@
...
+    const canvasElement = canvasRef?.current;
+    if (imageElement && canvasElement) {
+      canvasElement.width = imageElement.clientWidth;
+      canvasElement.height = imageElement.clientHeight;
+    }
+  };
+
+  useEffect(
+    () =>
+      autorun(() => {
```
isherman, line 53: check whether this is necessary

app/frontend/src/models/TarReader.ts
```diff
@@ -0,0 +1,132 @@
+const filenameMaxLength = 1024;
```
isherman, line 1: revisit

.vscode/react.code-snippets
```diff
@@ -0,0 +1,17 @@
+{
```
ethanrublee, line 1: this is cool, need to learn how to use that feature of vscode...

isherman:
Latest commit supports logs with unknown types in them. (I'll add the latest types too, just would like to try to ensure this doesn't have to update in lockstep with on-robot protos).

I did have to restructure the .tar file you attached a bit to conform with the path structure the scope currently expects. I should document somewhere.
[mycalibration01-ian.tar.zip](https://github.com/farm-ng/tractor/files/5278527/mycalibration01-ian.tar.zip)

ethanrublee (approved):
This works pretty well on my end.  I like it!  Merge at will.

isherman:
Thanks, going to merge this. Tracking TODOs in https://github.com/farm-ng/tractor/issues/106.

PR #105, opened 2020-09-23
Fix broken iOS Safari Navbar

isherman:
Also removes the "Debug" entry in the navbar, which doesn't do anything useful.


ethanrublee (approved):
Don't have iOS for testing, but changes look good.  Thanks for removing the Debug tab.

PR #110, opened 2020-09-27
Base pose camera calibration

ethanrublee:


ethanrublee:
Here is a log file created with commit 37df8f7

https://drive.google.com/file/d/1VjwxFR8PNqoFWRcq1OnBl84DYjhapndG/view?usp=sharing

To reproduce, 
Download above log and untar to /tmp/farm-ng-log

```bash
cd /tmp/farm-ng-log/
tar -xvzf ~/Downloads/base2camera_cal.tar.gz 
```

Then run the calibrator driver program
```bash
cd ~/code/tractor/build
make
cpp/farm_ng/calibration/base_to_camera_calibrator_driver --log /tmp/farm-ng-log/base2camera04/events-00821-00000.log --rig_calibration /tmp/farm-ng-log/base2camera04-resolve/apriltag_rig_model/solved-03269-00101.json  --out_archive base2camera_out_repro
```

You should see something like the following at the end of the run:
```
I0928 18:11:00.636502 24591 base_to_camera_calibrator_driver.cpp:278] root mean of residual error: 0.0119605
I0928 18:11:00.636528 24591 base_to_camera_calibrator_driver.cpp:174] base_pose_camera: position { x: -0.13741645251653897 y: 0.45764641200502704 z: 1.2687088991130309 } rotation { x: 0.49182646936492724 y: 0.49906452877987495 z: 0.50617753613303407 w: 0.502817682724109 } wheel_radius (inches): 8.09962 wheel_baseline (inches): 41.2541
I0928 18:11:00.656294 24591 base_to_camera_calibrator_driver.cpp:430] Wrote results to:
path: "base2camera_out_repro/base_to_camera/initial-24591-00000.pb" content_type: "application/protobuf; type=type.googleapis.com/farm_ng_proto.tractor.v1.BaseToCameraModel"
path: "base2camera_out_repro/base_to_camera/solved-24591-00001.pb" content_type: "application/protobuf; type=type.googleapis.com/farm_ng_proto.tractor.v1.BaseToCameraModel"
```

The tar ball also has a copy of the solved.pb to compare against.

cpp/farm_ng/tracking_camera.cpp
```diff
@@ -567,17 +568,21 @@ class TrackingCameraClient {
...
+            " x264enc bitrate=600 speed-preset=ultrafast tune=zerolatency ") +
+        "key-int-max=15 ! video/x-h264,profile=constrained-baseline ! queue " +
+        "max-size-time=100000000 ! h264parse ! ";
+
+    std::string encoder_omxh264 =
+        std::string(" omxh264enc control-rate=1 bitrate=1000000 ! ") +
+        " video/x-h264, stream-format=byte-stream !";
+
+    std::string cmd0 = std::string("appsrc !") + " videoconvert ! " +
+                       encoder_omxh264 +
```
ethanrublee, line 582: This should switch based on runtime or environment variable so that testing capture on desktop is possible.

python/farm_ng/canbus.py
```diff
@@ -93,7 +93,10 @@ def bind(self, interface):
     def send(self, cob_id, data, flags=0):
         cob_id = cob_id | flags
         can_pkt = struct.pack(self.FORMAT, cob_id, len(data), data)
-        self.sock.send(can_pkt)
+        try:
```
ethanrublee, line 96: Revert this.

python/farm_ng/steering.py
```diff
@@ -209,10 +212,13 @@ def send(self, n_periods):
                 if abs(velocity) >= 0.5:
                     velocity = np.sign(velocity) * (0.5/4 + (abs(velocity) - 0.5)*2)
                 self._target_speed = velocity
-                angular_velocity = np.clip(-self.joystick.get_axis_state('z', 0), -1.0, 1.0)*np.pi/3.0
-                self._command.angular_velocity = angular_velocity
+                self._target_angular_velocity = np.clip(-self.joystick.get_axis_state('z', 0), -1.0, 1.0)*np.pi/3.0
 
         self._command.velocity += np.clip((self._target_speed - self._command.velocity), -self._max_acc, self._max_acc)
+        self._command.angular_velocity += np.clip(
```
ethanrublee, line 218: @mmcgowen2015 this change adds angular acceleration to steering.

protos/farm_ng_proto/tractor/v1/calibrator.proto
```diff
@@ -57,7 +64,7 @@ message ApriltagRigTagStats {
   // If an RMSE is high on a particular image, it may be an outlier and this
   // granular information can be used for debugging issues in the calibration
   // data.
-  map<int32, double> per_image_rmse = 4;
+  repeated PerImageRmse per_image_rmse = 4;
```
ethanrublee, line 67: This change required because apparently map doesn't emmit proper json syntax from c++.

```
JSONDecodeError: Expecting ',' delimiter: line 324 column 9 (char 5424)
[1:21 PM]
"perImageRmse": [
    "22": 0.21919154720254469,
    "21": 0.19132221480022904,
```

cpp/farm_ng/tracking_camera.cpp
```diff
@@ -357,6 +357,7 @@ EventPb GenerateExtrinsicPoseEvent() {
   base_pose_t265.set_frame_b("tracking_camera/front/left");
   Sophus::SE3d se3(Sophus::SE3d::rotZ(-M_PI / 2.0));
```
ethanrublee, line 358: time to remove this, post merge i think.  Here we should load from the calibration if we have one.

cpp/farm_ng/calibrator.cpp
```diff
@@ -3,17 +3,15 @@
...
-#include <opencv2/imgproc.hpp>
-#include <opencv2/opencv.hpp>
-
-#include <ceres/ceres.h>
-#include <ceres/local_parameterization.h>
 #include <gflags/gflags.h>
 #include <glog/logging.h>
 #include <google/protobuf/util/json_util.h>
 
+#include "farm_ng/calibration/apriltag_rig_calibrator.h"
```
ethanrublee, line 10: most of the changes here are because I pulled out the apriltag rig calibration into the library, so this file deals with driving it from the eventbus.  May need to revisit this as we integrate with the frontend and robot programs.

cpp/farm_ng/calibration/apriltag_rig_calibrator.cpp
```diff
@@ -0,0 +1,502 @@
+#include "farm_ng/calibration/apriltag_rig_calibrator.h"
```
ethanrublee, line 1: this file is largely from the previous version of calibrator.

cpp/farm_ng/calibration/apriltag_rig_calibrator.cpp
```diff
@@ -0,0 +1,502 @@
...
+  }
+
+  LOG(INFO) << "root mean of residual error: "
+            << std::sqrt(summary.final_cost / summary.num_residuals);
+  ModelError(model);
+
+  return true;
+}
+
+Sophus::optional<NamedSE3Pose> EstimateCameraPoseRig(
```
ethanrublee, line 418: This function is new, it allows for finding the camera to an apriltag pose in one shot.

cpp/farm_ng/calibration/base_to_camera_calibrator_driver.cpp
```diff
@@ -0,0 +1,430 @@
+#include <fstream>
```
ethanrublee, line 1: This file is the meat of this pull request.

cpp/farm_ng/calibration/base_to_camera_calibrator_driver.cpp
```diff
@@ -0,0 +1,442 @@
...
+  }
+
+  virtual int GlobalSize() const { return size_; }
+
+  virtual int LocalSize() const { return size_; }
+  int size_;
+};
+
+template <typename T>
+static Sophus::SE3<T> TractorPoseDelta(
```
ethanrublee, line 84: Same equations used in: https://github.com/farm-ng/tractor/blob/f2eff899de3f39739d0c84683ad23fc6c5e29fe8/python/farm_ng/kinematics.py#L23

protos/farm_ng_proto/tractor/v1/calibrator.proto
```diff
@@ -100,6 +107,36 @@ message MonocularApriltagRigModel {
   repeated Image reprojection_images = 8;
 }
 
+message BaseToCameraModel {
+  message WheelState {
+    double wheel_velocity_rads_left = 1;
```
isherman, line 112: I feel like we typically are not including the unit in the variable / field name. Given it's in `WheelState`, could even be simplified to `velocity_left`

isherman: I also think `WheelMeasurement` might be a better name

protos/farm_ng_proto/tractor/v1/geometry.proto
```diff
@@ -45,6 +47,12 @@ message NamedSE3Pose {
   string frame_b = 3;
 }
 
+message Path {
```
isherman, line 50: - Should be a `Trajectory`, not a `Path`, right, if it includes a time parameterization?

- I'm wary of introducing the `Timestamp` field to `SE3Pose`. I see in other references, separate types for SE3 trajectory points that encapsulate a timestamp, vs. SE3 poses.

- Timestamp, from what I can tell, is intended to represent a calendar datetime since the unix epoch. I think we should consider using Duration, which could be relative to the start time of the trajectory.

https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/timestamp.proto
https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/duration.proto

protos/farm_ng_proto/tractor/v1/tractor.proto
```diff
@@ -10,16 +12,51 @@ message TractorState {
...
+  double commanded_wheel_velocity_rads_left = 5;
+  double commanded_wheel_velocity_rads_right = 6;
+
+  double target_unicycle_velocity = 7;
+  double target_unicycle_angular_velocity = 8;
+
+  // estimated odometry_pose_base, based on integrating the wheel velocity
+  // above, through the current kinematic model based on the TractorConfig.
+  // This is likely to slip based on poor traction moments, as its only based on
+  // motor hall effect tachometer measurements.  It may also be innacurate if
```
isherman, line 28: typo: "inaccurate"

protos/farm_ng_proto/tractor/v1/tractor.proto
```diff
@@ -10,16 +12,51 @@ message TractorState {
...
+  // the tractor config does not match the tractor's actual wheel baseline or
+  // radius.
+  NamedSE3Pose odometry_pose_base = 9;
+
+  // Timestamp of this estimate.
+  google.protobuf.Timestamp stamp = 10;
+
+  // The time duration between measurements, in seconds.
```
isherman, line 36: maybe more accurately, "between state updates"? are we integrating multiple measurements during this time duration?

protos/farm_ng_proto/tractor/v1/geometry.proto
ethanrublee: I like trajectory, will change.

I don't really see the risk with SE3Pose having an optionally set timestamp, it seems to make keeping the book keeping simple when copying poses from one structure to the next - i was trying to manage externally, but this felt like it was leading to more boiler plate and more risk of not setting the timestamp correctly.  For example, then we need StampedSE3Pose and NamedStampedSE3Pose.

I'm at the moment going to punt on using duration instead of time stamps, as we may need to compare timestamps across logs, and duration can always be computed by taking the difference.  The timestamp is now ubiquitous in the event bus, for stamping when things actually were measured in UTC time frame.

Discussed in person - let's punt on these to the design/<>.md files/discussion (create an issue to track), and gather more experience as we flesh out the applications.

ethanrublee: https://github.com/farm-ng/tractor/issues/112

protos/farm_ng_proto/tractor/v1/tractor.proto
ethanrublee: This is computed as the duration between state updates. Ideally the wheel velocities is averaged over this duration - at the moment the wheel velocities are just the latest sample from the canbus, which i'm not sure if this is instantaneous or filtered.  This is an area to improve upon, was trying to compute the average velocity based on the tachometer reading but this was giving inconsistent results.

cpp/farm_ng/calibration/apriltag_rig_calibrator.h
```diff
@@ -0,0 +1,76 @@
...
+  int NumFrames() const;
+
+  EventBus* bus_;
+  std::string rig_name_ = "rig";  // TODO(ethanrublee) set from command/config.
+  int root_id_;
+  std::unordered_set<int> ids_;
+  std::vector<ApriltagDetections> all_detections_;
+};  // namespace farm_ng
+
+bool Solve(ApriltagRigModel& model);
```
isherman, line 73: I'm surprised we have these nonmember functions (`Solve`, `ModelError`) in the global `farm_ng` namespace

cpp/farm_ng/calibration/base_to_camera_calibrator_driver.cpp
```diff
@@ -0,0 +1,430 @@
...
+          continue;
+        }
+        // LOG(INFO) << event.ShortDebugString();
+        Sophus::optional<NamedSE3Pose> o_camera_pose_rig =
+            farm_ng::EstimateCameraPoseRig(rig_model.rig(), detections);
+        if (o_camera_pose_rig) {
+          // set our event stamp as the pose stamp.
+          o_camera_pose_rig->mutable_a_pose_b()->mutable_stamp()->CopyFrom(
+              event.stamp());
+          if (!has_start && calibration_sample) {
```
isherman, line 377: we're pretty deep into nested `ifs` here, early return (or `continue` in this case) pattern might improve legibility

isherman (approved):

protos/farm_ng_proto/tractor/v1/calibrator.proto
ethanrublee: Ok, changed to WheelMeasurement.

protos/farm_ng_proto/tractor/v1/geometry.proto
ethanrublee: Changed to TrajectorySE3.

cpp/farm_ng/calibration/apriltag_rig_calibrator.h
ethanrublee: I need to fix that up, but not in this PR. I'm still workshopping the idiomatic interface for modelling calibration and solving it.

cpp/farm_ng/calibration/base_to_camera_calibrator_driver.cpp
ethanrublee: Agree, this is an ugly statemachine for parsing the sequence messages in the log.  Not sure how to improve without rewriting. Added a continue.

Suggest this is revisited when we add the event_bus driven protocol.

ethanrublee:
Collecting one more dataset and testing that it runs on the robot before merge.

ethanrublee:
Captured another dataset and ran on tractor0003.

```
I0929 15:34:22.226161 20032 base_to_camera_calibrator_driver.cpp:272] root mean of residual error: 0.0472945
I0929 15:34:22.226227 20032 base_to_camera_calibrator_driver.cpp:162] base_pose_camera: position { x: -0.10016184836937912 y: 0.37664054348385578 z: 8.8378529868470341 } rotation { x: 0.50860399303821036 y: 0.48046078748300869 z: 0.49
624330579722303 w: 0.51402528284923787 } wheel_radius (inches): 7.73276 wheel_baseline (inches): 37.9881
```

Need to collect a few more datasets and refine the optimization a bit more in follow ons.  It looks like the Z is not very well constrained which makes sense - a hand measurement or estimation of the ground plane from tags or feature points could help.  Also, may be safest to hand measure the wheel radius and wheel baseline, and keep these fixed.

Still pleased with the result.

Download log from here:
https://drive.google.com/file/d/18U2IvyhJSfreOyfpviYCktJEA2TvHKYL/view?usp=sharing

To reproduce:
```bash
cd /tmp/farm-ng-log/
tar -xvzf ~/Downloads/base2cal_2020_09_29_01.tar.gz 
~/code/tractor/build/cpp/farm_ng/calibration/base_to_camera_calibrator_driver --log /tmp/farm-ng-log/base2cal_2020_09_29_01/events-07514-00000.log --rig_calibration /tmp/farm-ng-log/base2cal_2020_09_29_01/apriltag_rig_model/solved-07935-00101.json --out_archive base2cal_2020_09_29_01_repro
```
```

ethanrublee:
Calibration fixture used in test:

PXL_20200929_195010275
PXL_20200929_195053510

And a video of capturing the above dataset: https://youtu.be/bVsSQKW0mWw

Issue #117, opened 2020-09-30
Allow for initial values to be provided for base to camera calibration


ethanrublee:
At the moment, there are several parameters that we would like good initial guesses for, and possibly to hold some of these constant:

- wheel radius
- wheel baseline
- rotation of camera
- x,y,z of camera

Rotation would be nice if there were several defaults, like "front facing, upside down", or "front facing", "rear facing", "rear facing, upside down".

A diagram would help when filling out a form, to prevent confusion.

ethanrublee:
@isherman this is another one i think to consider in the programs work.

PR #118, opened 2020-10-01
Base2cam fixes

ethanrublee:
This fixes a crash if there is a tag detection which is not in the rig.  Also add command line parameters for setting the wheel radius and wheel baseline, and viewing direction, as some of our cameras are right side up versus upside down.  This now converges on tractor0002 and tractor0003 data.


ethanrublee:
This partially adds support for #117

isherman (approved):

PR #124, opened 2020-10-07
Capture video dataset

ethanrublee:


cpp/farm_ng/tracking_camera.cpp
```diff
@@ -668,6 +709,46 @@ class TrackingCameraClient {
     }
   }
 
+  void record_every_frame(cv::Mat image, google::protobuf::Timestamp stamp) {
+    if (!latest_command_.has_record_start()) {
+      frame_video_writer_->Close();
+      return;
+    }
+    if (latest_command_.record_start().mode() !=
```
isherman, line 717: a little weird to me that these methods are always invoked, and have the responsibility of checking `latest_command_` (rather than checking at the callsite)

isherman (approved):
Approvied (one minor comment): https://github.com/farm-ng/tractor/pull/124#discussion_r501933415

isherman (approved):

PR #128, opened 2020-10-09
(work in progress) Webrtc signaling server

jinfwhuang:
Add signaling-server that could be act as an alternative frontend entrypoint. 

- No breaking changes to proxy-server, `go/webrtc/cmd/proxy-server/main.go`.
- If proxy-server is started with a signaling server endpoint configured, e.g. `-signal=http://192.168.1.137:8586`, the browser application served by the signaling-server would be able to discover/connect to the proxy-server.
- A signaling-server could be deployed outside of the local network.


isherman:
nice! I was able to reproduce when connecting to the signaling server you provided, with both tractor and laptop on my local network. but not when I used my phone connected to LTE (as a way to test NAT traversal). it looks like we don't have STUN servers enabled? that might be all that's necessary to get NAT traversal working. happy to work on it together.

jinfwhuang:
Updated to add Stun server for frontend <-> proxy communication.

PR #129, opened 2020-10-09
Visual tracking 01

ethanrublee:


python/farm_ng/ipc.py
```diff
@@ -28,6 +28,9 @@
 
 
 def host_is_local(hostname, port):
+    #logger.info(hostname)
```
ethanrublee, line 31: I found we're still calling this evilness.. It causes the python event loop to block for ~10 seconds when running in AP mode.

app/frontend/src/components/scope/visualizers/CaptureVideoDatasetConfiguration.tsx
```diff
@@ -31,6 +31,17 @@ const CaptureVideoDatasetConfigurationForm: React.FC<FormProps<
           setValue((v) => ({ ...v, name }));
         }}
       />
+      <Form.Group
+        label="Detect Apriltags?"
```
ethanrublee, line 35: @isherman Seems pretty easy to add form fields! Nice job.

cpp/farm_ng/calibration/CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ endforeach()
 add_library(farm-ng-calibration
   ${cpp_files}
   ${h_files}
+  time_series.h
```
ethanrublee, line 19: TODO add tests, add to cpp_files, h_files as above.

app/frontend/src/components/scope/visualizers/CaptureVideoDatasetConfiguration.tsx
isherman: oh good! unfortunately you inadvertently stumbled into a minor landmine (a lego piece on the floor?) that my find-and-replace missed. Because React uses [event pooling](https://reactjs.org/docs/events.html#event-pooling), it's unsafe to refer to the event `e` in an asynchronous callback like the one we pass to `setValue()` -- it may be undefined by the time it's dereferenced. You'll see in all the other non-checkbox callbacks, we explicitly save `e.target.value` into another variable... it should be the same here.

ethanrublee:
Some initial 3d reconstruction based on this tracker: https://www.youtube.com/watch?v=6vXuIq2NvIc

It appears to basically close a 200ft square loop (only using local features, so there is some drift, looks mostly in z).  Too early, haven't quite quantified it.  But this means the optimizer is converging, and we have a starting point for servoing based on natural features.

I'm using a static mask to help the tracker exclude points on the tractor and in the sky.

tracking_mask

ethanrublee:
Some examples based off the current code +- a few minor changes: https://www.youtube.com/playlist?list=PLWQmpzk0y9NDT3lz8A9Vanddrht62wQte

ethanrublee:
Some field tests:
Pass 1:
https://youtu.be/jNxFge7oJ6I
Pass 2:
https://youtu.be/9AXb1dOELuI
Pass 3:
https://youtu.be/vb_xGnq3GHU

ethanrublee:
https://youtu.be/TczKtRQaTmE

app/frontend/src/components/scope/visualizers/CaptureVideoDatasetConfiguration.tsx
isherman: just pushed a change to this branch with a fix

cpp/farm_ng/calibration/epipolar.h
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// https://web.stanford.edu/class/cs231a/course_notes/03-epipolar-geometry.pdf
+//
+// Given b_pose_a, point_b = b_pose_a*point_a
+// Returns Essential matrix (b_E_a) such that:
+//
+//   point_b.transpose().dot(b_E_a.dot(point_a)) == 0
+template <typename Scalar>
+Eigen::Matrix<Scalar, 3, 3> EssentionalMatrix(
```
isherman, line 18: typo: `Essential`

cpp/farm_ng/calibration/flow_book_keeper.cpp
```diff
@@ -0,0 +1,391 @@
...
+      flow_window_(21, 21),
+      flow_max_levels_(3) {
+  cv::RNG rng;
+  for (int i = 0; i < 1000; ++i) {
+    colors_.push_back(cv::Scalar(rng.uniform(0, 255), rng.uniform(0, 255),
+                                 rng.uniform(0, 255)));
+  }
+  int image_width = camera_model_.image_width();
+  int image_height = camera_model_.image_height();
+  std::string mask_path =
```
isherman, line 28: should be able to use `GetBucketRelativePath(BUCKET_CONFIGURATIONS)`

cpp/farm_ng/calibration/flow_book_keeper.cpp
```diff
@@ -0,0 +1,391 @@
...
+    int crowd_window = 31;
+    cv::rectangle(mask,
+                  cv::Rect(pt.x - crowd_window / 2, pt.y - crowd_window / 2,
+                           crowd_window, crowd_window),
+                  cv::Scalar::all(0), -1);
+    GenFlowPoint(flow_image, kpt.pt);
+  }
+  LOG(INFO) << "keypoints.size() after filter " << fast_count;
+
+  return;
```
isherman, line 312: unreachable code below?

cpp/farm_ng/calibration/flow_book_keeper.cpp
```diff
@@ -0,0 +1,391 @@
...
+uint64_t FlowBookKeeper::LastImageId() const {
+  CHECK_GT(image_id_gen_, 0);
+  return image_id_gen_ - 1;
+}
+
+const FlowImage* FlowBookKeeper::PreviousFlowImage() const {
+  if (flow_images_.empty()) {
+    return nullptr;
+  }
+  CHECK(image_id_gen_ > 0);
```
isherman, line 350: here and elsewhere in the file, could be replaced with a call to `LastImageId` that already CHECKs

cpp/farm_ng/calibration/flow_book_keeper.h
```diff
@@ -0,0 +1,108 @@
...
+#include <sophus/se3.hpp>
+
+#include "farm_ng/calibration/time_series.h"
+
+#include "farm_ng_proto/tractor/v1/camera_model.pb.h"
+
+namespace farm_ng {
+using farm_ng_proto::tractor::v1::CameraModel;
+
+struct FlowPointImage {
```
isherman, line 18: For consistency with `FlowPointWorld`, `id` and `point_image`?

cpp/farm_ng/calibration/flow_book_keeper.h
```diff
@@ -0,0 +1,108 @@
...
+  uint64_t flow_point_id;
+  Eigen::Vector2f image_coordinates;
+};
+
+class FlowPointWorld {
+ public:
```
isherman, line 24: struct?

cpp/farm_ng/calibration/flow_book_keeper.h
```diff
@@ -0,0 +1,108 @@
...
+  uint64_t id;
+
+  Eigen::Vector3d point_world;
+  // set because we want it ordered.
+  std::set<uint64_t> image_ids;
+  double rmse = 0;
+};
+
+class FlowImage {
+ public:
```
isherman, line 34: struct?

cpp/farm_ng/calibration/flow_book_keeper.h
```diff
@@ -0,0 +1,108 @@
...
+  void RemovePointWorld(uint64_t id);
+  void RemoveFlowImage(uint64_t id);
+
+  uint64_t LastImageId() const;
+
+  const FlowImage* EarliestFlowImage() const;
+  const FlowImage* PreviousFlowImage() const;
+  FlowImage* MutablePreviousFlowImage();
+
+  FlowImage* MutableFlowImage(uint64_t id);
```
isherman, line 72: i notice this is a pointer even though internal CHECKs mean that it will always return non-null. i see conflicting advice on best practices around this question of API design -- curious how you think about it.

cpp/farm_ng/calibration/time_series.h
```diff
@@ -0,0 +1,77 @@
...
+  static bool Compare(const ValueT& lhs, const ValueT& rhs) {
+    return lhs.stamp() < rhs.stamp();
+  }
+
+  bool empty() const { return series_.empty(); }
+  const_iterator lower_bound(const google::protobuf::Timestamp& stamp) const {
+    ValueT x;
+    x.mutable_stamp()->CopyFrom(stamp);
+    return std::lower_bound(series_.begin(), series_.end(), x,
+                            [](const ValueT& lhs, const ValueT& rhs) {
```
isherman, line 24: could you use `Compare` instead of this lambda?

cpp/farm_ng/calibration/flow_book_keeper.h
ethanrublee: Ok will try.

cpp/farm_ng/calibration/flow_book_keeper.h
ethanrublee: I don't want these to be accidentally copied, and these pointers are copied into containers for bundle adjustment; kind of following the convention of protobuf.  Not entirely happy with it.  Google style and c++ idiomatic approaches seem to conflict here.

cpp/farm_ng/calibration/time_series.h
ethanrublee: probably.

cpp/farm_ng/tracking_camera.cpp
```diff
@@ -639,7 +581,29 @@ class TrackingCameraClient {
...
+         std::string("^logger/.*"),
+         // subscribe to steering commands to toggle new goals for VO
+         std::string("^steering$"),
+         // tracking camera commands, recording, etc.
+         std::string("^tracking_camera/command$"),
+         // tractor states for VO
+         std::string("^tractor_state$")});
+
+    auto base_to_camera_path =
+        GetBlobstoreRoot() / "base_to_camera_models/base_to_camera.json";
```
isherman, line 596: also available via `GetBucketRelativePath(BUCKET_BASE_TO_CAMERA_MODELS)`

cpp/farm_ng/tracking_camera.cpp
```diff
@@ -639,7 +581,29 @@ class TrackingCameraClient {
...
+    if (boost::filesystem::exists(base_to_camera_path)) {
```
isherman, line 597: note to self -- could be provided by a blobstore API that returns a std::optional

cpp/farm_ng/tracking_camera.cpp
```diff
@@ -757,6 +739,18 @@ class TrackingCameraClient {
...
 
+  void RecordEveryApriltagFrame(cv::Mat image,
+                                google::protobuf::Timestamp stamp) {
+    auto apriltags = detector_->Detect(image, stamp);
+    auto image_pb = frame_video_writer_->AddFrame(image, stamp);
+
+    apriltags.mutable_image()->CopyFrom(image_pb);
+    event_bus_.Send(farm_ng::MakeEvent(
+        "calibrator/tracking_camera/front/apriltags", apriltags, stamp));
+    event_bus_.Send(farm_ng::MakeEvent("tracking_camera/front/apriltags",
```
isherman, line 750: note to self: just thinking about event naming drift.. i wonder if ipc logger should log the current git commit as metadata

cpp/farm_ng/calibration/epipolar.h
ethanrublee: Epic.

protos/farm_ng_proto/tractor/v1/capture_video_dataset.proto
```diff
@@ -7,6 +7,7 @@ option go_package = "github.com/farm_ng/genproto";
 
 message CaptureVideoDatasetConfiguration {
   string name = 1;
+  bool detect_apriltags = 2;
```
isherman, line 10: could be a `TrackingCameraCommand_RecordStart_Mode` (not sure i actually like that idea)

third_party/install.sh
```diff
@@ -2,7 +2,7 @@
 
 . ../setup.bash
 
-./install-opencv.sh
+#./install-opencv.sh
```
isherman, line 5: maybe a TODO

isherman (approved):
minor stuff; approved; your judgement on what feedback you want to take

protos/farm_ng_proto/tractor/v1/capture_video_dataset.proto
ethanrublee: This seemed clearer, to translate it in the program to the mode.

third_party/install.sh
ethanrublee: let's revisit this soon, in the effort to Dockerize.

PR #132, opened 2020-10-12
Load tractor config and apriltag config from blobstore

isherman:


cpp/farm_ng/tracking_camera.cpp
```diff
@@ -363,6 +362,18 @@ Sophus::SE3d ApriltagPoseToSE3d(const apriltag_pose_t& pose) {
...
+double TagSize(const TagLibrary& tag_library, int tag_id) {
+  auto it = std::find_if(tag_library.tags().begin(), tag_library.tags().end(),
+                         [tag_id](const TagConfig& tag_config) {
+                           return tag_config.id() == tag_id;
+                         });
+
+  if (it != tag_library.tags().end()) {
+    return it->size();
+  }
+  return 0;
```
ethanrublee, line 374: should probably be non-zero, or we should handle zero well in the code (such as not estimating the pose or populating the size).  Or default could be 1.0.

protos/farm_ng_proto/tractor/v1/tractor.proto
```diff
@@ -59,4 +59,14 @@ message TractorConfig {
 
   // Extrinsic relationships to sensors.
   repeated NamedSE3Pose base_poses_sensor = 7;
+
+  message DiffDrive {
```
ethanrublee, line 63: We will have 3 variants:

2 wheel diff drive, 4 wheel skid drive, 2 track drive (4 motors).  Will think on some ways to encode this.  Might be nice to share some structures here, in tractorstate, and the basetocamera model, as this data layer becomes more use case constrained.

if we wish to punt, suggest adding an enum for drive type, and we'll assume the wheel radius is constant across all wheels.

cpp/farm_ng/calibration/apriltag_rig_calibrator.cpp
```diff
@@ -343,6 +343,7 @@ void ApriltagRigCalibrator::AddFrame(ApriltagDetections detections) {
   for (auto it = detections.mutable_detections()->begin();
        it != detections.mutable_detections()->end();) {
     if (it->tag_size() == 0.0) {
+      // TODO: What's the correct behavior here?
```
isherman, line 346: @ethanrublee thoughts?

cpp/farm_ng/tracking_camera.cpp
```diff
@@ -816,8 +838,6 @@ class TrackingCameraClient {
 void Cleanup(farm_ng::EventBus& bus) {}
 
 int Main(farm_ng::EventBus& bus) {
-  // Declare RealSense pipeline, encapsulating the actual device and sensors
-  rs2::pipeline pipe;
```
isherman, line 820: @ethanrublee can you confirm this was a typo? (it's also initialized as a TrackingCameraClient member variable)

protos/farm_ng_proto/tractor/v1/tractor.proto
isherman: went with the enum from the PR you merged yesterday

cpp/farm_ng/tracking_camera.cpp
isherman: now defaults to 1.0

isherman:
I think this is ready for review.

One issue I ran into is that google protobuf libraries, when serializing messages to json/text, don't include empty nested message fields (even with `including_default_value_fields=True`). So as long as we have nested messages in our configuration files, `gentractor`, `genapriltag` and similar utilities won't print the full message .

Workarounds I can think of:
- Hardcode in initialization of those nested messages
- Don't use nested messages in configuration protos
- Write our own spec-compatible json/text serialization (spent a little bit of time on this, felt like a rabbit hole)

In our conversation we discussed having `tracking_camera` (re)load the TagLibrary when it switches into a capture mode. Is that still desirable, and if so do you think it's important to include in this PR?

cpp/farm_ng/calibration/apriltag_rig_calibrator.cpp
ethanrublee: oh sorry i think i lost the response i was working on here. will take another look.

cpp/farm_ng/blobstore.cpp
```diff
@@ -21,11 +12,15 @@ boost::filesystem::path GetBlobstoreRoot() {
   return root;
 }
 
-boost::filesystem::path GetBucketRelativePath(BucketId id) {
-  return kBucketPaths.at(id);
+boost::filesystem::path GetBucketRelativePath(Bucket id) {
+  std::string path =
+      Bucket_Name(id).substr(std::strlen("BUCKET_"), std::string::npos);
+  std::transform(path.begin(), path.end(), path.begin(),
```
ethanrublee, line 18: can you just pass std::tolower here instead of the lambda?

cpp/farm_ng/calibration/apriltag_rig_calibrator.cpp
ethanrublee: here I think the tag should be excluded similar to if the ``ids_.count(id) == 0`` case below.

ethanrublee: Perhaps a warning message if the ids_.count(id) != 0 but we don't have a size.

cpp/farm_ng/tracking_camera.cpp
ethanrublee: Looks like it!

python/farm_ng/tractor.py
```diff
@@ -72,8 +72,7 @@ def __init__(self, event_bus):
         self.odom_poses_tractor = TimeSeries(1.0)
         self.odom_pose_tractor = SE3.identity()
 
-        # TODO(ethanrublee|isherman) load from disk somehow.
-        self.config = default_config()
+        self.config = TractorConfig.saved()
```
ethanrublee, line 75: We should use the config for the 4 wheel drive switch below, rather than the presence of a file.

ethanrublee:
I feel like empty nested messages aren't a big deal if we are typically editing with a script or the web interface.

It would be nice for tracking_camera to reload the apriltag library on record_start, so we don't need it in a follow up.

Would like tractor.py to load 4 wheel drive from config, rather than testing HERE file.

ethanrublee (approved):
Looks good, approved, but suggest making changes in comments above.

cpp/farm_ng/calibration/apriltag_rig_calibrator.cpp
isherman: 👍 done

cpp/farm_ng/blobstore.cpp
isherman: done

isherman:
> It would be nice for tracking_camera to reload the apriltag library on record_start, so we don't need it in a follow up.

done

PR #134, opened 2020-10-14
Disable proto-lint on ARM64

isherman:


ethanrublee (approved):

isherman:
@ethanrublee simplified this; looking for another approval just so you have eyes on it

PR #136, opened 2020-10-14
Add subscriptions to IPC

### LINKED ISSUE #70: event_bus is slow when we have a few processes sending lots of messages
Implement message filtering on send side, based on processes which seem interested in a given topic.

When running the tracking_camera process, which sends 200hz pose + imu data, all processes receive this data.  On the nano, the decoding negatively impacts the steering and tractor process (doubling the CPU utilization...).

isherman:
Closes #70


python/farm_ng/pose_vis_toy.py
```diff
@@ -1,35 +1,33 @@
-import asyncio
 import logging
 import sys
 
 import numpy as np
+from farm_ng_proto.tractor.v1.geometry_pb2 import NamedSE3Pose
```
ethanrublee, line 5: We can probably remove this file.

cpp/farm_ng/extrinsic_calibrator.cpp
```diff
@@ -21,6 +21,10 @@ class ExtrinsicCalibrator {
   ExtrinsicCalibrator(boost::asio::io_service& io_service)
       : bus_(GetEventBus(io_service, "extrinsic-calibrator")),
         timer_(io_service) {
+    Subscription subscription;
+    // TODO: What events does this service need?
```
isherman, line 25: @ethanrublee currently some of our programs (this one, tracking_camera, ipc_logger) rely on the `event.data.type_url()` to filter messages they care about) -- e.g. [here](https://github.com/farm-ng/tractor/pull/136/files#diff-d80da11ed7c016238cd360d234d52aa71160b3a71b576953e2f9c504df379d4aR62). can we express an event name filter for them? or should i implement filtering based on type, in addition to name?

python/farm_ng/pose_vis_toy.py
isherman: Current our `test-integration.sh` and the associated Dockerfile rely on it. Should I remove that stuff too?

cpp/farm_ng/extrinsic_calibrator.cpp
ethanrublee: We can probably use name for most filters, and assume a one to one name to type relationship.

python/farm_ng/pose_vis_toy.py
ethanrublee: I would be in favor of rewriting those, or disabling them for now and revisiting integration testing, it doesn't really seem to cover much at the moment.

cpp/farm_ng/extrinsic_calibrator.cpp
ethanrublee: This file can be removed.

cpp/farm_ng/ipc.cpp
```diff
@@ -265,6 +282,27 @@ class EventBusImpl {
...
+      const Announce& announce = it.second;
+      if (std::any_of(
+              announce.subscriptions().begin(), announce.subscriptions().end(),
+              [event](const Subscription& subscription) {
+                std::smatch match;
+                return std::regex_search(
+                    event.name(), match,
+                    // TODO:
+                    // https://stackoverflow.com/questions/6846510/how-to-convert-string-to-regex-literal
+                    std::regex(subscription.name()));
```
ethanrublee, line 298: Should this be memoized? not sure how expensive regex compilation is.

cpp/farm_ng/tracking_camera.cpp
```diff
@@ -610,6 +609,11 @@ class TrackingCameraClient {
       : io_service_(bus.get_io_service()), event_bus_(bus) {
     event_bus_.GetEventSignal()->connect(std::bind(
         &TrackingCameraClient::on_event, this, std::placeholders::_1));
+
+    Subscription subscription;
+    subscription.set_name(".*");
```
ethanrublee, line 614: This could be more strict it seems.

isherman:
As expected, `tractor.py` on this branch **with** T265 pose streaming looks like `tractor.py` on master **with no** T265 pose streaming ('cause it doesn't receive any poses).
Screen Shot 2020-10-14 at 10 19 07 PM

cpp/farm_ng/ipc.cpp
```diff
@@ -265,6 +282,34 @@ class EventBusImpl {
...
  private:
+  const std::vector<boost::asio::ip::udp::endpoint> recipients(
+      const Event& event) {
+    std::vector<boost::asio::ip::udp::endpoint> result;
+    for (auto& it : recv_.announcements()) {
+      const Announce& announce = it.second;
+      if (std::any_of(
+              announce.subscriptions().begin(), announce.subscriptions().end(),
+              [event](const Subscription& subscription) {
+                // TODO: memoize
```
ethanrublee, line 293: One idea: it would be neat if recipients could be tracked through a member variable ``unordered_map<EventName, std::vector<endpoint> >``, which is updated anytime a new event name is sent (not tracked in the map), or the list of subscriptions changes.  The return type from this function could then be a ``const&``.  This would reduce the overhead of this lookup significantly.

cpp/farm_ng/ipc_logger.cpp
```diff
@@ -20,6 +21,9 @@ class IpcLogger {
         log_writer_(nullptr),
         log_timer_(bus_.get_io_service()),
         announce_timer_(bus_.get_io_service()) {
+    Subscription subscription;
```
ethanrublee, line 24: I wonder if this should be passed in through the Main() function in ``init.h`` as this add subscription seems fairly boiler plate.

python/farm_ng/motor.py
```diff
@@ -144,9 +144,10 @@ def __init__(
         self.gear_ratio = gear_ratio
         self.poll_pairs = poll_pairs
         self.max_current = 20
+        self._event_bus = get_event_bus(self.name)
         self._latest_state = motor_pb2.MotorControllerState()
         self._latest_stamp = Timestamp()
-        self.can_socket.add_reader(self._handle_can_message)
+        # self.can_socket.add_reader(self._handle_can_message)
```
ethanrublee, line 150: maybe we should just add a environment variable or flag to enable mocking this (noop is fine in the moc).

ethanrublee:
The performance improvements look great.  And implementation is nice and straightforward.  Love it!

cpp/farm_ng/ipc_logger.cpp
isherman: I like keeping the logic in the program constructor, I think it co-locates the concerns a bit more closely, and allows the program to easily use `event_bus_.name()` in a subscription if it wants to. The primary motivation for making `AddSubscriptions` the API was so that sub-objects could register additional subscriptions.

But I did add an overloaded method signature to reduce boilerplate -- it's a one-liner now.

cpp/farm_ng/ipc.cpp
isherman: I spent some time thinking about this, but I didn't like the idea of making an `announcements_` setter concerned about invalidating a `recipients_` cache. I also couldn't really think of a way to detect when subscriptions had changed without defining a custom hash function. As a compromise, just caching the compiled regexes for now. Definitely open to implementing as you suggest if you can share some C++-fu.

cpp/farm_ng/tracking_camera.cpp
isherman: fixed

cpp/farm_ng/ipc.cpp
ethanrublee: I think for now this is fine, and is more of an optimization that we could do if we discover a lot of time spent in send and recipient matching.  Mainly concerned that the cost of sending is number of services * number of subscriptions with the current impl in all languages, where it could be just number of subscriptions that match.

cpp/farm_ng/ipc_logger.cpp
ethanrublee: Sounds good.

isherman:
@ethanrublee tested locally; ready for review

cpp/farm_ng/capture_video_dataset.cpp
```diff
@@ -43,6 +44,9 @@ class CaptureVideoDatasetProgram {
     } else {
       set_configuration(configuration);
     }
+    bus_.AddSubscriptions(
+        {bus_.GetName(), "tracking_camera/front/left/image", "logger/status"});
```
ethanrublee, line 48: should this be ``"^"+bus_.GetName()``?

cpp/farm_ng/capture_calibration_dataset.cpp
```diff
@@ -37,6 +39,8 @@ class CaptureCalibrationDatasetProgram {
     } else {
       set_configuration(configuration);
     }
+    bus_.AddSubscriptions(
+        std::vector<std::string>{"^calibrator/", "logger/status"});
```
ethanrublee, line 43: suggest making syntax consistent across AddSubscriptions call.  Here using the ``vector<string>{}``, other places just ``{}``

cpp/farm_ng/ipc.cpp
```diff
@@ -21,8 +22,11 @@
 #include "farm_ng_proto/tractor/v1/tracking_camera.pb.h"
 
 namespace farm_ng {
+using farm_ng_proto::tractor::v1::Announce;
+using farm_ng_proto::tractor::v1::Event;
 using farm_ng_proto::tractor::v1::LoggingCommand;
 using farm_ng_proto::tractor::v1::LoggingStatus;
+using farm_ng_proto::tractor::v1::Subscription;
 using farm_ng_proto::tractor::v1::TrackingCameraCommand;
```
ethanrublee, line 30: For future, would be great to remove ``tracking_camera.pb.h`` dependency from ``ipc.h/cpp``

cpp/farm_ng/tracking_camera.cpp
```diff
@@ -610,6 +609,9 @@ class TrackingCameraClient {
       : io_service_(bus.get_io_service()), event_bus_(bus) {
     event_bus_.GetEventSignal()->connect(std::bind(
         &TrackingCameraClient::on_event, this, std::placeholders::_1));
+
+    event_bus_.AddSubscriptions({"^tracking_camera/command$"});
```
ethanrublee, line 613: Perhaps we should have a comment string documenting the regex intent?  Like here is this meant to be exact match of "tracking_camera/command", "^" starts with, and "$" end of line?

go/webrtc/cmd/proxy-server/main.go
```diff
@@ -41,6 +41,9 @@ func main() {
 		Channel:              eventChan,
 		PublishAnnouncements: true,
 	})
+	// Currently we forward all eventbus traffic over the webRTC proxy.
+	// TODO: Allow webRTC peers to dynamically adjust subscriptions
+	eventBus.AddSubscriptions([]string{".*"})
```
ethanrublee, line 46: Was thinking one way would be to publish the list of event names in the announce messages.

python/farm_ng/tractor.py
```diff
@@ -54,19 +54,19 @@ def find_nearest(self, stamp):
 
 
 class TractorController:
-    def __init__(self, event_bus):
-        self.event_loop = asyncio.get_event_loop()
+    def __init__(self, event_bus: EventBus):
```
ethanrublee, line 57: Thanks for adding the type specification.

ethanrublee (approved):
Some minor comments.  I haven't tested yet on my tractor, should probably do that before merge.

ethanrublee:
Ok, was able to build and teleop the tractor, captured a video dataset.

ethanrublee:
merge at will.

isherman:
Good feedback -- let's get this in and iterate on subscription naming / documentation / API affordances as we start using.

PR #137, opened 2020-10-16
Unify vscode / pre-commit python import formatting

isherman:
I was getting conflicts between my VSCode and pre-commit because they relied on two different import ordering tools. Should be unified now.


ethanrublee (approved):
Thank you for fixing.  Haven't tried your change, but I was experiencing the same conflict!

PR #138, opened 2020-10-18
Add clang-tidy

isherman:
I'm a fan of opinionated tooling to enforce style. I've been spoiled by languages that do have good built-in tooling to help with this -- it just makes my brain hurt less, makes code review easier, helps new developers get up to speed. I've been playing with getting some additional tooling in place for C++ (at the minimum, clang-tidy and iwyu... in the future tsan, asan, additional static analyzers, etc. )

There are a number of ways this could be configured (custom CMake targets, integration via `compile_commands.json`, etc) that we should probably discuss. This PR takes the following approach:
 - In `pre-commit` (and CI), run `clang-tidy` over all `cpp/farm_ng` files whenever a `.cpp` file changes
 - Disable all clang-tidy rules that our codebase currently violates. We can "ratchet" towards removing most of these over time.
 - For help fixing clang-tidy issues , add a build variant `clang-tidy` in vscode which will display clang-tidy issues as warning squiggles. This avoids adding the overhead of `clang-tidy` to our default builds.  

Screen Shot 2020-10-18 at 3 53 50 PM

Screen Shot 2020-10-18 at 3 44 36 PM

Unfortunately I don't think we get auto-fixes, using the Intellisense that's bundled with the default `ms-vscode.cpptools` extension. I got auto-fixes working with [this alternative clangd-based extension](https://marketplace.visualstudio.com/items?itemName=llvm-vs-code-extensions.vscode-clangd), but I don't think it's worth dealing with another CPP extension.

### DIFF
--- a/bootstrap-apt.sh
+++ b/bootstrap-apt.sh
@@ -35,6 +35,8 @@ sudo apt-get install -y \
      build-essential \
      ca-certificates \
      clang \
+     clang-10 \
+     clang-tidy-10 \
      cmake \
      curl \
      dirmngr \
@@ -46,6 +48,7 @@ sudo apt-get install -y \
      libboost-filesystem-dev \
      libboost-regex-dev \
      libboost-system-dev \
+     libclang-10-dev \
      libeigen3-dev \
      libgoogle-glog-dev \
      libgstreamer-plugins-base1.0-dev \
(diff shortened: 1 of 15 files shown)


ethanrublee (approved):
Thanks!

bootstrap-apt.sh
```diff
@@ -35,6 +35,8 @@ sudo apt-get install -y \
      build-essential \
      ca-certificates \
      clang \
+     clang-10 \
```
ethanrublee, line 38: Have you verified this on the nano?

isherman: yup, works on the nano

isherman:
struggling to get to a reasonable build times in CI. 
clang-tidy relies on `compile_commands.json`; cmake requires both system dependencies and third_party to be installed.
curious if you can think of a way to avoid those installation steps... otherwise i'm thinking this might require a docker image

ethanrublee:
docker might be the way to go, or just deal with long CI times for c++ changes.

ethanrublee:
For docker and CI i want to move towards something similar to what i sketched at : https://github.com/farm-ng/farm_ng_calibration

ethanrublee:
Here the system deps are installed quite quickly (2 minutes): https://github.com/farm-ng/farm_ng_calibration/runs/1121569354

That uses a combination of debs and conda packages (conda packages for prebuilt c++ dependencies and python packages).  Might be a somewhat heavy lift for this PR, not sure if its worth it yet.  Perhaps a compromise is just adding this to local pre-commit and then we can figure out CI later.

PR #145, opened 2020-10-26
Fixes make targets

jinfwhuang:
Move the building of protos and frontend applications inside a docker. Two advantages:
- No need to install yarn and nodejs in dev box or nano
- Remove potential bash version requirements for these two make commands

-------------
I encountered these two bash issues in my macos build.

1. Re: `make protos`
```
# This setting causes an error on older version of bash
shopt -s globstar
```

2. Re `make frontend`
```
$ echo $BASH_VERSION
5.0.18(1)-release

$ cp -rT app/frontend/dist build/frontend
cp: illegal option -- T
usage: cp [-R [-H | -L | -P]] [-fi | -n] [-apvXc] source_file target_file
       cp [-R [-H | -L | -P]] [-fi | -n] [-apvXc] source_file ... target_directory
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,3 +43,6 @@ build/
 
 # virtual env
 env/
+
+# yarn
+app/frontend/.cache
--- a/Makefile
+++ b/Makefile
@@ -20,8 +20,7 @@ cpp:
 	cd build && rm -rf ./* && cmake -DCMAKE_PREFIX_PATH=`pwd`/../env -DCMAKE_BUILD_TYPE=Release .. && make -j`nproc --ignore=1`
 
 frontend:
-	cd app/frontend && yarn && yarn build
-	cp -rT app/frontend/dist build/frontend
+	scripts/build-frontend.sh
 
 protos:
 	scripts/build-protos.sh
--- a/scripts/Dockerfile.protoc
+++ b/scripts/Dockerfile.protoc
@@ -2,6 +2,8 @@ FROM golang:alpine
 # redo
 # Install protoc
 RUN apk add --update-cache \
+  bash \
+  findutils \
   git \
   nodejs-current-npm \
   protobuf \
--- a/scripts/build-frontend.sh
+++ b/scripts/build-frontend.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+set -e
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+CMD="bash"
+CMD_ARGS="scripts/scripts_in_docker/build-frontend.sh"
+TAG="node"
+
+docker run \
+       --rm -v $PWD:/src:rw,Z -u $(id -u):$(id -g) --workdir /src \
+       --entrypoint $CMD $TAG $CMD_ARGS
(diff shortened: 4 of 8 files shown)


ethanrublee:
@jinfwhuang thank you for the improvements.  A recent set of changes adapter our CI system to build docker images for deployment and integration testing.  One of the TODOs is to now port the development workflow to leverage the docker images somehow.

Will come back to this PR during that development. @isherman

isherman:
Yes, thanks @jinfwhuang @ethanrublee. I'd definitely like to dockerize the development environment more, the development environment is too hard to replicate / reproduce currently (see for example https://github.com/farm-ng/tractor/issues/154).

Sorry @jinfwhuang, there's been some churn in the `protoc` workflow in the last couple weeks that probably makes this PR mostly stale. 

But if this is still a pain point for you I'd be in favor of a PR that dockerized at least enough of the build to reduce your pain.

Some references that could be useful:
- https://github.com/farm-ng/tractor/blob/master/Makefile#L27
    - ^ all protoc codegen is now handled by cmake. `make protos` just builds the go and ts codegen targets.
- https://github.com/isherman/tractor/blob/ef94bba7e379e6789a6e37e05a85564cbdb5173a/dev.Dockerfile
    - ^ I've been using this as a base image in which I can test the bootstrap scripts. I launch it and run `make`.
- https://github.com/farm-ng/tractor/blob/master/docker/base.Dockerfile#L64
    - ^ another potentially useful base image. i'd like to remove the code duplication between this and `bootstrap-apt.sh`

jinfwhuang:
Sure. I think the fixes in this PR is more or less deprecated b/c of your latest changes.

I put a few words of caution about going down the route of leveraging too much docker on building artifacts. Using docker in the build tools constrains the outputs to be useful for one specific runtime/arch. If that is the goal, it is usually easier to just hardcode all the parameters and requirements for that specific target. For example, the simpler approach would just say that the build tooling works if this os + toolchain (bash, go, python, clang) are such and such. It is also hard to link things up when individual steps are dockerized, and linking build steps within docker is very clumsy. IDE supports also becomes an issue; but you might not care.

I have not seen much success in using docker in build toolings. It works fine for the first few targets, and once the target interactions get complex, it gets hard to manage. I have worked on two repos where that route is tried and have to be abandoned after some time. It is possible that this repo never gets to that scale.

isherman:
> Sure. I think the fixes in this PR is more or less deprecated b/c of your latest changes.
> 
> I put a few words of caution about going down the route of leveraging too much docker on building artifacts. Using docker in the build tools constrains the outputs to be useful for one specific runtime/arch. If that is the goal, it is usually easier to just hardcode all the parameters and requirements for that specific target. For example, the simpler approach would just say that the build tooling works if this os + toolchain (bash, go, python, clang) are such and such. It is also hard to link things up when individual steps are dockerized, and linking build steps within docker is very clumsy. IDE supports also becomes an issue; but you might not care.
> 
> I have not seen much success in using docker in build toolings. It works fine for the first few targets, and once the target interactions get complex, it gets hard to manage. I have worked on two repos where that route is tried and have to be abandoned after some time. It is possible that this repo never gets to that scale.

OK, that's really useful perspective. We won't take any immediate action in that area. Going to close this one.

PR #147, opened 2020-10-27
N cameras

ethanrublee:


isherman (approved):

cpp/farm_ng/calibration/apriltag_rig_calibrator.cpp
```diff
@@ -179,10 +192,16 @@ void ModelError(ApriltagRigModel& model) {
                    -1);
       }
     }
+    cv::imshow("reprojection", image);
```
ethanrublee, line 195: TODO remove.

docker/cpp.Dockerfile
```diff
@@ -0,0 +1,63 @@
+FROM ubuntu:18.04
```
ethanrublee, line 1: I guess these should be removed, bad merge? @isherman

isherman: done

cpp/farm_ng/capture_video_dataset.cpp
```diff
@@ -137,6 +137,7 @@ class CaptureVideoDatasetProgram {
   }
 
   void on_event(const EventPb& event) {
+    // TODO(isherman): Discuss right filter with @ethanrublee
```
isherman, line 140: @ethanrublee what's the right filter to use here? should it read the `camera.json` configuration?

isherman: - rely on camera.json
- calculate per-camera count
- search for other "tracking_camera/front" strings

isherman: See 1089c3aea97530613a3dd4aad4a77209bd7d8132

PR #157, opened 2020-11-22
allow user to override source controlled vs code settings

jinfwhuang:
Use `.vscode/workspace.code-workspace` instead of `.vscode/settings.json` to keep track of shared vscode settings.


.vscode/workspace.code-workspace
```diff
@@ -0,0 +1,151 @@
+{
+  "folders": [
+    {
+      "path": ".."
+    }
+  ],
+  "settings": {
+    "clang-format.style": "google",
```
jinfwhuang, line 8: The original content of `settings.json` is move here.

.vscode/sample-settings.json
```diff
@@ -0,0 +1,18 @@
+{
```
jinfwhuang, line 1: A local copy of `settings.json` could be used to override what is already been set in `workspace.code-workspace`

modules/perception/cpp/farm_ng/perception/frame_grabber_intel.cpp
```diff
@@ -123,6 +123,7 @@ class FrameGrabberIntel : public FrameGrabber {
       std::stringstream ss;
       ss << "RealSense error calling " << e.get_failed_function() << "("
          << e.get_failed_args() << "):\n    " << e.what();
+      LOG(ERROR) << ss.str();
```
jinfwhuang, line 126: Where a runtime_error was throw, I am getting an error message that looks like the following. It doesn't provide sufficient debugging information. 
```
*** Aborted at 1606030777 (unix time) try "date -d @1606030777" if you are using GNU date ***
PC: @                0x0 (unknown)
*** SIGABRT (@0x3db8) received by PID 15800 (TID 0x7f86f53450) from PID 15800; stack trace: ***
    @       0x7f8a2856c0 ([vdso]+0x6bf)
    @       0x7f8930e4f8 raise
Aborted (core dumped)
```

This error happened when I didn't properly configure the camera config file to have the properly serial number.

isherman (approved):

isherman:
cool, thanks. so we'll want to "open workspace" in vscode now, instead of "open folder", right? another nice thing about the workspace file is that I can add additional folders to it -- for example, the go language plugins only work properly if you're working in a workspace folder with `go.mod`, so my workaround is `Add Folder > modules/frontend/go` (https://github.com/golang/go/issues/32394)

jinfwhuang:
Yup. "open workspace" is how I would open the project. 

With the workspace setting control, one could also have multiple `.vscode/settings.json` file to control different folders.

modules/perception/cpp/farm_ng/perception/frame_grabber_intel.cpp
ethanrublee: Thanks, yeah looks like the error handing scheme globally needs some work.  Not sure what's causing the SIGABRT.

For the framegrabber in particular, some verbosity, such as listing the available cameras might be worthwhile.

jinfwhuang: I googled around that error and tested a tiny bit. It is related to using glog. It hides the actual error somehow. I didn't really look into it.

PR #159, opened 2020-11-29
Intrinsic calibration

ethanrublee:


modules/core/cpp/farm_ng/core/init.cpp
```diff
@@ -53,14 +53,14 @@ void GlogFailureFunction() {
...
          void (*cleanup_func)(EventBus&)) {
-  gflags::ParseCommandLineFlags(&argc, &argv, true);
   FLAGS_logtostderr = 1;
-
   std::string filename = boost::filesystem::path(argv[0]).filename().string();
   google::InitGoogleLogging(filename.c_str());
   google::InstallFailureFunction(&GlogFailureFunction);
   google::InstallFailureSignalHandler();
 
+  gflags::ParseCommandLineFlags(&argc, &argv, true);
```
ethanrublee, line 62: TODO remove this, a bit of thrash.

modules/perception/cpp/farm_ng/capture_video_dataset.cpp
```diff
@@ -53,8 +53,8 @@ class CaptureVideoDatasetProgram {
     } else {
       set_configuration(configuration);
     }
-    bus_.AddSubscriptions(
-        {bus_.GetName(), "/image$", "/apriltags$", "logger/status"});
+    bus_.AddSubscriptions({bus_.GetName(), "/image$", "/apriltags$",
+                           "logger/command", "logger/status"});
```
ethanrublee, line 57: Adding logger/command seems necessary for the archive to work as expected, where writing a file is in the directory of the current log.

modules/perception/cpp/farm_ng/create_video_dataset.cpp
```diff
@@ -0,0 +1,310 @@
...
+            << "Image should be grayscale or BGR, something is wrong...";
+        gray = image.clone();
+      }
+      auto stamp = core::MakeTimestampNow();
+      bus_.Send(
+          MakeEvent(camera_model->frame_name() + "/image", image_pb, stamp));
+
+      status_.set_num_frames(image_pb.frame_number().value() + 1);
+      bool first_frame_for_camera = true;
+      for (auto& entry : *status_.mutable_per_camera_num_frames()) {
```
ethanrublee, line 141: This frame/tag detection code is duplicated from capture_video_dataset, should likely be pulled into a library.

modules/perception/cpp/farm_ng/perception/pose_graph.h
```diff
@@ -0,0 +1,365 @@
...
+    out << self.frame_a << " <- " << self.frame_b;
+    if (self.a_pose_b) {
+      out << " t: " << self.a_pose_b->translation().transpose()
+          << " r:" << self.a_pose_b->unit_quaternion().vec().transpose();
+    }
+    return out;
+  }
+};
+
+class PoseGraph {
```
ethanrublee, line 90: Most of this implementation can go in c++ file.  Also, review api... add a few unit tests.

modules/perception/protos/farm_ng/perception/capture_video_dataset.proto
```diff
@@ -6,9 +6,16 @@ import "google/protobuf/timestamp.proto";
...
 
+message VideoFileCamera {
+  string camera_frame_name = 1;
+  farm_ng.core.Resource video_file_resource = 2;
+}
+
 message CaptureVideoDatasetConfiguration {
   string name = 1;
   bool detect_apriltags = 2;
+  repeated VideoFileCamera video_file_cameras = 3;
```
ethanrublee, line 17: Not sure if we should add ``CreateVideoDataset*`` or support both use cases from this one configuration.

ethanrublee: One reason I reused this is that the calibration routines take as input a CaptureVideoDatasetResult as their input,

bootstrap.sh
```diff
@@ -9,5 +9,10 @@ while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symli
 done
 export FARM_NG_ROOT=$( cd "$( dirname "${SOURCE}" )" >/dev/null 2>&1 && pwd )
 
+FARM_NG_GOPATH=$FARM_NG_ROOT/env/go
```
isherman, line 12: this duplicates some logic in `https://github.com/farm-ng/tractor/blob/master/bootstrap-venv.sh#L21`; it seems slightly closer to the context where it's used there, but i'm fine with moving it here.

ethanrublee: I seemed to still have issues, but maybe a red-herring.

modules/calibration/cpp/farm_ng/calibration/CMakeLists.txt
```diff
@@ -18,7 +19,12 @@ add_library(farm_ng_calibration
   ${h_files}
 )
 target_include_directories(farm_ng_calibration PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/../..)
-target_link_libraries(farm_ng_calibration farm_ng_calibration_protobuf farm_ng_perception ${CERES_LIBRARIES})
+target_link_libraries(farm_ng_calibration
+farm_ng_calibration_protobuf
+farm_ng_perception
+${CERES_LIBRARIES}
+${GLOG_LIBRARIES}
```
isherman, line 26: good to start thinking about how changes like this could affect downstream projects relying on our tarball releases (with their own build system)

modules/calibration/protos/farm_ng/calibration/calibrate_intrinsics.proto
```diff
@@ -0,0 +1,45 @@
...
+package farm_ng.calibration;
+option go_package = "github.com/farm-ng/genproto/calibration";
+
+message CalibrateIntrinsicsConfiguration {
+  // Serialized CaptureVideoDatasetResult
+  farm_ng.core.Resource video_dataset = 1;
+
+  farm_ng.perception.CameraModel.DistortionModel distortion_model = 2;
+
+  // A list of Apriltag board names to use in calibration.
```
isherman, line 18: do we have the concept of a calibration board in our codebase? is this just an apriltag rig?

modules/calibration/protos/farm_ng/calibration/intrinsic_model.proto
```diff
@@ -0,0 +1,51 @@
...
+
+
+message IntrinsicModel {
+  // The solved for camera model.
+  farm_ng.perception.CameraModel camera_model = 1;
+
+  // Calibration board rig used for solving camera model
+  repeated farm_ng.perception.ApriltagRig apriltag_rigs = 2;
+
+  // Is this the initial rig, or did the solver converge, fail...
```
isherman, line 23: Check comments throughout -- lots of references to "rigs" that I think are just copy/paste artifacts

modules/calibration/protos/farm_ng/calibration/calibrate_intrinsics.proto
```diff
@@ -0,0 +1,45 @@
...
+  // Serialized CalibrateIntrinsicsResult
+  farm_ng.core.Resource result = 2;
+
+  SolverIterationSummary iteration_summary = 3;
+}
+
+message CalibrateIntrinsicsResult {
+  // Configuration which initiated the computation of this result.
+  CalibrateIntrinsicsConfiguration configuration = 1;
+  farm_ng.core.Resource intrinsics_solved = 2;
```
isherman, line 39: `// Serialized IntrinsicModel`

modules/calibration/protos/farm_ng/calibration/intrinsic_model.proto
```diff
@@ -0,0 +1,51 @@
...
+import "farm_ng/perception/image.proto";
+import "farm_ng/calibration/calibrator.proto";
+import "google/protobuf/timestamp.proto";
+
+package farm_ng.calibration;
+option go_package = "github.com/farm-ng/genproto/calibration";
+
+
+
+message IntrinsicModel {
```
isherman, line 16: i notice this is in a separate file, where the other models are in `calibrator.proto` -- aim to split them all out?

modules/perception/protos/farm_ng/perception/capture_video_dataset.proto
isherman: If there's only one program that runs to capture / create, then I like a single configuration.
If they are two different programs, I don't think they should share the same configuration message.

ethanrublee: But what about the result?  It seems like quite a bit of extra boiler plate to have a separate protobuf per program.

modules/calibration/cpp/farm_ng/calibration/CMakeLists.txt
ethanrublee: I was hitting issues with glog/gflag being linked inconsistently, I think this can be removed now that the CERES is built as a shared library.  For other build systems that are hermetic this shouldn't be an issue.

modules/calibration/protos/farm_ng/calibration/calibrate_intrinsics.proto
ethanrublee: It is just an apriltag rig.

ethanrublee: Need to copy the apriltag rigs from the configuration directory into the blobstore for this to work.  https://github.com/farm-ng/tractor/tree/master/modules/calibration/config/calibration_boards

modules/calibration/protos/farm_ng/calibration/intrinsic_model.proto
ethanrublee: Yes i'd like to split them all out.

modules/perception/cpp/farm_ng/capture_video_dataset.cpp
isherman: That's right, I ran into this recently as well.

modules/perception/cpp/farm_ng/create_video_dataset.cpp
```diff
@@ -0,0 +1,289 @@
...
+        }
+      }
+
+      bus_.get_io_service().poll();
+
+      image_pb.mutable_frame_number()->set_value(
+          image_pb.frame_number().value() + 1);
+    }
+
+    bus_.get_io_service().poll();
```
isherman, line 185: could we collapse this line and L179 by making it the first statement of the `while` loop?

modules/perception/cpp/farm_ng/create_video_dataset.cpp
```diff
@@ -0,0 +1,289 @@
...
+    result.mutable_configuration()->CopyFrom(configuration_);
+    result.mutable_per_camera_num_frames()->CopyFrom(
+        status_.per_camera_num_frames());
+    result.mutable_per_tag_id_num_frames()->CopyFrom(
+        status_.per_tag_id_num_frames());
+
+    result.mutable_stamp_end()->CopyFrom(MakeTimestampNow());
+    result.mutable_dataset()->set_path(log.recording().archive_path());
+
+    // TODO some how save the result in the archive directory as well, so its
```
isherman, line 196: isn't that what's happening here?

modules/perception/cpp/farm_ng/perception/apriltag.h
```diff
@@ -71,6 +74,45 @@ class ApriltagDetector {
...
+      insert(rig, node);
+    }
+  }
+
+  std::optional<FrameNames> GetFrameNames(int tag_id) {
+    auto it_name = id_frame_map_.find(tag_id);
+    if (it_name != id_frame_map_.end()) {
+      return it_name->second;
+    }
+    return std::optional<FrameNames>();
```
isherman, line 102: `std::nullopt`

modules/perception/protos/farm_ng/perception/create_video_dataset.proto
```diff
@@ -0,0 +1,57 @@
...
+  string name = 1;
+  bool detect_apriltags = 2;
+  repeated VideoFileCamera video_file_cameras = 3;
+  repeated farm_ng.core.Resource apriltag_rigs = 4;
+}
+
+
+message CreateVideoDatasetStatus {
+  // Number of frames captured per camera
+  repeated PerCameraNumFrames per_camera_num_frames = 4;
```
isherman, line 26: surprising field numbers

modules/perception/cpp/farm_ng/perception/pose_graph.h
isherman: I have some nits about the naming, for example: `AverageAPoseB()` currently invokes `GetAPoseB()`, which happens to return an average.

Either the "average" nature of the query is a property of the PoseGraph -- in which case `GetAPoseB` always returns an average, and `APoseB()` will return an average. Or it's a property of the query  -- in which case `AverageAPoseB()` should invoke `GetAverageAPoseB()`. I don't think the API currently makes it clear which of the two it is.

But doesn't need to block this PR

isherman (approved):

PR #168, opened 2020-12-12
First pass at Depthmap type and using it in the Image proto.

ethanrublee:
- [x] Read depth from Image blob and write to disk as png in capture_robot_extrinsics_dataset.cpp
- [x] Adds support for joint space requests in the robot HAL.
For later PR TODO 
- if depth is available evaluate its use in calibration losses and solution stats.
- Add sample code for encoding/decoding depthmap in python/c++


modules/perception/cpp/farm_ng/perception/frame_grabber_k4a.cpp
```diff
@@ -0,0 +1,177 @@
...
+        k4a::transformation depth_to_color(calibration);
+        dev_->start_cameras(&config);
+        while (true) {
+          // Poll the device for new image data.
+          //
+          // We set the timeout to 0 so we don't block if there isn't an
+          // available frame.
+          //
+          // This works here because we're doing the work on the same thread
+          // that we're using for the UI, and the ViewerWindow class caps the
```
isherman, line 96: Is this true / does this involve a GUI?

isherman (approved):

modules/perception/cpp/farm_ng/perception/frame_grabber_k4a.cpp
ethanrublee: No this is cruft from copying the example from the SDK.  Will clean up.

ethanrublee: Changed the timeout to 100 milliseconds.

Also, this code may need revisiting after a bit of use.

PR #169, opened 2020-12-12
Add detect_apriltags to detect tags on a video dataset. 

ethanrublee:
…t on some existing datasets.

### DIFF
--- a/modules/calibration/cpp/farm_ng/calibration/multi_view_apriltag_rig_calibrator.cpp
+++ b/modules/calibration/cpp/farm_ng/calibration/multi_view_apriltag_rig_calibrator.cpp
@@ -319,13 +319,6 @@ std::vector<MultiViewApriltagDetections> LoadMultiViewApriltagDetections(
 
   std::map<std::string, TimeSeries<Event>> apriltag_series;
 
-  std::map<std::string, std::shared_ptr<perception::ApriltagDetector>> per_camera_detectors;
-  std::optional<perception::ApriltagConfig> tag_config;
-  if(config.has_tag_config()) {
-  tag_config = ReadProtobufFromResource<perception::ApriltagConfig>(config.tag_config());
-
-  }
-  ImageLoader image_loader;
   while (true) {
     EventPb event;
     try {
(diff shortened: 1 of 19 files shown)


ethanrublee (commented):
Ready for review.

modules/calibration/cpp/farm_ng/calibration/multi_view_apriltag_rig_calibrator.cpp
```diff
@@ -349,14 +354,15 @@ std::vector<MultiViewApriltagDetections> LoadMultiViewApriltagDetections(
   ApriltagsFilter tag_filter;
   std::vector<MultiViewApriltagDetections> mv_detections_series;
   auto time_window =
-      google::protobuf::util::TimeUtil::MillisecondsToDuration(1000.0 / 7);
+      google::protobuf::util::TimeUtil::MillisecondsToDuration(1000.0 / 2);
```
isherman, line 357: You're confident about this as a general change?

modules/frontend/python/farm_ng/frontend/program_supervisor.py
```diff
@@ -48,6 +48,12 @@
         name='Capture Video Dataset',
         description='Capture video segments, for use in other programs',
     ),
+    'detect_apriltags': ProgramInfo(
```
isherman, line 51: I'd like to load these from a `configurations/programs.json` in the future

modules/perception/cpp/farm_ng/detect_apriltags.cpp
```diff
@@ -0,0 +1,278 @@
...
+      }
+      bus_.Send(event);
+
+      Image image;
+      ApriltagDetections unfiltered_detections;
+      bool has_detections = false;
+      if (event.data().UnpackTo(&image)) {
+        if (image.camera_model().distortion_coefficients_size() == 0) {
+          for (int i = 0; i < 8; ++i) {
+            image.mutable_camera_model()->add_distortion_coefficients(0);
```
isherman, line 116: Is this the right place to fill these, or should downstream programs support camera models that are missing distortion coefficients?

isherman (approved):

modules/perception/cpp/farm_ng/detect_apriltags.cpp
ethanrublee: That's a good question, probably downstream use needs to be robust to zero length coeffs.

modules/calibration/cpp/farm_ng/calibration/multi_view_apriltag_rig_calibrator.cpp
ethanrublee: I'm not, will need to revisit this timewindow.  We choose the closest event in the window, but this may be too forgiving, and hide systematic capture bugs.

PR #177, opened 2020-12-18
Add integration test structure

isherman:
[Installing a self hosted runner](https://docs.github.com/en/free-pro-team@latest/actions/hosting-your-own-runners/about-self-hosted-runners)

A few things on my mind here:
- Support for multiple types of runners with different hardware resources: "onsite", "onsensorbox", "ontractor", etc.
- Module decoupling
- Support for multiple runners on the same machine (e.g. don't use host networking)
- Support for tests in parallel within a job (use Python's asyncio)
- Test result reporting (use Python's unittest library)
- Integration test structure
   a. populate a blobstore
   b. run a group of processes w/ docker-compose 
   c. inspect the blobstore


isherman:
TODO: 
- [ ] Improve speed
- [ ] More integration tests
- [ ] Fix pre-commit issue "Duplicate module named 'integration_test'"
- [ ] Pull dockerhub `edge` images rather than building them

ethanrublee:
From a security perspective,  how does the runner limit itself to only the master branch?  Or should we actually point the runners to a private "fork" of the repository, so that pull requests can't change the workflow configuration and execute code on the runner?

isherman:
> From a security perspective, how does the runner limit itself to only the master branch? Or should we actually point the runners to a private "fork" of the repository, so that pull requests can't change the workflow configuration and execute code on the runner?

Ack, you're right, I misunderstood how Github handles edits to `.github/workflows` in PRs. I didn't think it trusted them but I guess it does. Let me look into the private "fork" idea.

isherman:
OK, scoped this back, with the plan to tackle non-host-mode networking in a future PR. 

The test definition is still pretty verbose. Curious if you see any ways to simplify it.

ethanrublee (approved):

PR #179, opened 2020-12-22
Attempt to optimize docker build.

ethanrublee:


docker/devel.Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+RUN add-apt-repository "deb http://realsense-hw-public.s3.amazonaws.com/Debian/apt-repo bionic main" -u
+RUN apt-get update && apt-get install -y --no-install-recommends librealsense2-dev librealsense2-utils
+
+# Install third-party python
+COPY setup.bash env.sh requirements.txt ./
+RUN pip3 install virtualenv
+RUN virtualenv ./env
+RUN . ./env/bin/activate && pip install -r ./requirements.txt
+
+# Build third-party c++
```
isherman, line 54: Maybe move this before third-party python, as it's likely to take a lot longer

isherman (approved):

docker/devel.Dockerfile
ethanrublee: I'll do that next time its changed, as this took a while to build even on my machine.

PR #180, opened 2020-12-22
Add UI for capture_robot_extrinsics_dataset, log_playback

isherman:


modules/calibration/protos/farm_ng/calibration/capture_robot_extrinsics_dataset.proto
```diff
@@ -46,25 +47,19 @@ message CaptureRobotExtrinsicsDatasetConfiguration {
...
   }
   // Serialized CaptureRobotExtrinsicsDatasetResult.
   farm_ng.core.Resource result = 2;
 
   // Configuration which initiated this program.
   CaptureRobotExtrinsicsDatasetConfiguration configuration = 3;
 
-  // The pending queue of requests
-  repeated CapturePoseRequest request_queue = 4;
+  // DEPRECATED
```
isherman, line 61: Deprecated because it's fully described by the configuration (now that poses are explicit in the config).

modules/calibration/cpp/farm_ng/capture_robot_extrinsics_dataset.cpp
```diff
@@ -245,12 +258,8 @@ class CaptureRobotExtrinsicsDatasetProgram {
 }  // namespace farm_ng::calibration
 
 int Main(farm_ng::core::EventBus& bus) {
-  auto configuration = ReadProtobufFromJsonFile<
-      farm_ng::calibration::CaptureRobotExtrinsicsDatasetConfiguration>(
-      farm_ng::core::GetBlobstoreRoot() / FLAGS_configuration_path);
-
   farm_ng::calibration::CaptureRobotExtrinsicsDatasetProgram program(
-      bus, configuration);
+      bus, FLAGS_configuration_path, FLAGS_interactive);
```
isherman, line 262: The pattern is slightly different here than other programs. The frontend program boilerplate is different too. We could keep the boilerplate consistent by making the `CaptureRobotExtrinsicsDatasetConfiguration` just contain a `Resource`, since that's what the CLI and frontend actually expect. Or find a more general pattern, that lets any `Configuration` be specified via a Resource path.

modules/frontend/frontend/src/components/scope/visualizers/ApriltagRig.tsx
```diff
@@ -57,7 +57,8 @@ const ApriltagRig3DElement: React.FC<SingleElementVisualizerProps<
     value: [, value],
   } = props;
 
-  const { nodes, rootTagId } = value;
+  const { nodes } = value;
+  const rootTagId = value.rootTagId || nodes[0].id;
```
isherman, line 61: discuss

modules/frontend/frontend/src/components/scope/visualizers/CaptureRobotExtrinsicsDatasetConfiguration.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+  const {
+    name,
+    workspaceFrameName,
+    baseFrameName,
+    linkFrameName,
+    halServiceAddress,
+  } = value;
+
+  console.log(value);
```
isherman, line 31: remove

modules/calibration/cpp/farm_ng/capture_robot_extrinsics_dataset.cpp
ethanrublee: I feel like that's a bit indirect for the Config to have resource that points to another Config.  Suggest supporting a duality - configuration can be populated by a resource or config, through the use of the input_required  oneof convention, as you have it now.

modules/calibration/protos/farm_ng/calibration/capture_robot_extrinsics_dataset.proto
ethanrublee: K makes sense, should we just remove this field?

isherman: suggest we leave it in, as we captured a dataset that contained it.

isherman: similarly above (L52)

cmake/farm_ng_modules.cmake
```diff
@@ -77,7 +77,7 @@ macro(farm_ng_add_protobufs target)
       VERBATIM)
 
     # ts
-    set("_protoc_args_ts" "--ts_proto_out=forceLong=long:${_proto_output_dir_ts}")
+    set("_protoc_args_ts" "--ts_proto_out=forceLong=long,outputClientImpl=false:${_proto_output_dir_ts}")
```
ethanrublee, line 80: Add a note here, why its false for future.

isherman: done

modules/frontend/frontend/src/components/scope/visualizers/ApriltagRig.tsx
isherman: @ethanrublee is this a reasonable interpretation of a `rootTagId` equal to 0, that we just choose an arbitrary node as the root?

modules/calibration/cpp/farm_ng/capture_robot_extrinsics_dataset.cpp
isherman: makes sense, i like it.

modules/frontend/frontend/src/components/scope/visualizers/ApriltagRig.tsx
ethanrublee: I think so.

ethanrublee:
Some notes from testing this locally:

- Could the 3d view go below the images in the Result display?
- 3d view is quite large and not very interesting at the moment - suggest making it smaller somehow, and adding the base reference frame marker.  From a calibration perceptive it would be nice to see the cloud of poses
- The joint plotting is currently text only.  I think a 2d plot would be more engaging, but this is optional.
- Huh, the program result is not in the log, so doesn't come through on scope.  And i'd like to be able to browse to the blobstore and view the result, but that appears to give a blank page, and this error in the console:
- ```react-dom.production.min.js:209 TypeError: xe is not a function or its return value is not iterable
    at uPlot.esm.js:2296
    at Array.forEach (<anonymous>)
    at uPlot.esm.js:2270
    at Oe (uPlot.esm.js:2470)
    at Te (uPlot.esm.js:2512)
    at Le (uPlot.esm.js:2594)
    at re (uPlot.esm.js:1740)
    at ft (uPlot.esm.js:3262)
    at new ip (uPlot.esm.js:3275)
    at Plot.tsx:33
es @ react-dom.production.min.js:209
```

**Comment by isherman** [comment]
> Some notes from testing this locally:
> 
> * Could the 3d view go below the images in the Result display?

Done

> * 3d view is quite large and not very interesting at the moment - suggest making it smaller somehow, and adding the base reference frame marker.  From a calibration perceptive it would be nice to see the cloud of poses

Done

> * The joint plotting is currently text only.  I think a 2d plot would be more engaging, but this is optional.

Done

> * Huh, the program result is not in the log, so doesn't come through on scope.  And i'd like to be able to browse to the blobstore and view the result, but that appears to give a blank page, and this error in the console:
> * ```
>     at uPlot.esm.js:2296
>     at Array.forEach (<anonymous>)
>     at uPlot.esm.js:2270
>     at Oe (uPlot.esm.js:2470)
>     at Te (uPlot.esm.js:2512)
>     at Le (uPlot.esm.js:2594)
>     at re (uPlot.esm.js:1740)
>     at ft (uPlot.esm.js:3262)
>     at new ip (uPlot.esm.js:3275)
>     at Plot.tsx:33
>   ```
> 
> es @ react-dom.production.min.js:209

This was a bug, should be fixed now.

The latest set of changes actually changed a bunch of stuff to incorporate this (very reasonable) feedback. Would you mind reproducing on your end again?

**Comment by ethanrublee** [comment]
Will try to reproduce soon. Maybe tonight if I have a chance.

**Comment by ethanrublee** [comment]
Love it!

File: modules/core/cpp/farm_ng/log_playback.cpp [hunk 6]
```diff
@@ -41,12 +45,59 @@ std::chrono::microseconds ToChronoDuration(
...
+      return;
+    }
+    status_timer_.expires_from_now(std::chrono::milliseconds(1000));
+    status_timer_.async_wait(std::bind(&IpcLogPlayback::on_status_timer, this,
+                                       std::placeholders::_1));
+
+    send_status();
+  }
+
+  void on_event(const EventPb& event) {
```
ethanrublee, line 88: Wish list for future playback:

- Prefix messages with an optional namespace, so it doesn't conflict with other messages on a live system
- Add timeline, scrub, pause, step, etc.

ethanrublee (approved):

PR #183, opened 2020-12-23
WIP: bag file to video

collinbrake:


modules/perception/cpp/farm_ng/rs2_bag_to_eventlog.cpp
```diff
@@ -0,0 +1,149 @@
...
+  VideoFromBagProgram(EventBus& bus,
+                            VideoFromBagConfiguration configuration)
+                            //,bool interactive)
+      : bus_(bus), timer_(bus_.get_io_service()) {
+    if (interactive) {
+      status_.mutable_input_required_configuration()->CopyFrom(configuration);
+    } else {
+      set_configuration(configuration);
+    }
+    bus_.AddSubscriptions({bus_.GetName(), "logger/command", "logger/status"});
```
ethanrublee, line 39: Here, i think we only need the {bus_.GetName()} subscription.

modules/perception/protos/farm_ng/perception/rs2_bag_to_eventlog.proto
```diff
@@ -0,0 +1,65 @@
...
+
+package farm_ng.perception;
+option go_package = "github.com/farm-ng/genproto/perception";
+
+message BagFileCamera {
+  string camera_frame_name = 1;
+  //farm_ng.core.Resource video_file_resource = 2;
+}
+
+message VideoFromBagConfiguration {
```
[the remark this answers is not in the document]
collinbrake: It seems like EventLog is used in other cases, not Eventlog. Should probably add an underscore in the file name too, since event and log aren't one word.

modules/perception/cpp/farm_ng/perception/intel_bag_utils.h
```diff
@@ -0,0 +1,67 @@
+#ifndef FARM_NG_INTEL_BAG_UTILS_H_
+#define FARM_NG_INTEL_BAG_UTILS_H_
+#include <librealsense2/rsutil.h>
+#include <librealsense2/rs.hpp>
+#include <opencv2/imgproc.hpp>
+
+namespace farm_ng {
+namespace perception {
+
+namespace {
```
ethanrublee, line 10: Remove the anonymous namespace from the header.

modules/perception/cpp/farm_ng/perception/intel_bag_utils.h
```diff
@@ -0,0 +1,67 @@
...
+
+// Convert rs2::frame to cv::Mat
+// Supported frame formats include BGR, RGB, depth maps, and disparity
+// Any RGB frame is converted to BGR format
+// https://raw.githubusercontent.com/IntelRealSense/librealsense/master/wrappers/opencv/cv-helpers.hpp
+cv::Mat RS2FrameToMat(const rs2::frame& f) {
```
ethanrublee, line 16: Suggest keeping implementation in a cpp file, as this will result in multiple definitions (unless using inline) and slower compile times.

modules/perception/cpp/farm_ng/perception/frame_grabber_intel.cpp
```diff
@@ -1,4 +1,5 @@
 #include "farm_ng/perception/frame_grabber.h"
+#include "farm_ng/perception/intel_bag_utils.h"
```
ethanrublee, line 2: suggest name like, ``intel_rs2_utils.h``.

modules/perception/cpp/farm_ng/rs2_bag_to_event_log.cpp
```diff
@@ -0,0 +1,182 @@
...
+      // signal.
+      cv::imshow("Color", color);
+      char c = static_cast<char>(cv::waitKey(1));
+      if (tolower(c) == 'q')
+      {
+         cout << "Quit signal recieved, stopping video.\n\n";
+         break;
+      }
+
+      std::optional<CameraModel> camera_model;
```
ethanrublee, line 110: If this is set every frame, it doesn't need to be optional.  Or it can be set the first time around the loop, via the optional.

modules/perception/cpp/farm_ng/rs2_bag_to_event_log.cpp
```diff
@@ -0,0 +1,182 @@
...
+        image_pb.mutable_camera_model()->CopyFrom(*camera_model);
+      }
+
+      CHECK_EQ(camera_model->image_width(), w);
+      CHECK_EQ(camera_model->image_height(), h);
+
+      // Copy the data to the image protobuf, TODO
+      // How do we actualy serialize the image?
+      // Cannot find this in create_video_dataset.
+      //image_pb.mutable_resource->CopyFrom(...)
```
ethanrublee, line 125: Suggest using the VideoStreamer here.

```c++
// define this at the top of the loop, constructed once you get the first camera_model.
std::unique_ptr<VideoStreamer> streamer;

// inside loop once camera_model is set
if(camera_model && !streamer) { 
  streamer = std::make_unique<VideoStreamer>(
    bus_, 
    *camera_model,
    // later we will replace MODE_MP4_FILE with MODE_JPG_SEQUENCE
    // but that's not implemented yet
    VideoStreamer::MODE_MP4_FILE); 
} 
// note stamp here should be from the stamp of the frame structure grabbed from the bag
auto stamp = google::protobuf::util::TimeUtil::MillisecondsToTimestamp(
          color_frame.get_timestamp());
Image image_pb = streamer->AddImage(color, stamp);

File: modules/perception/cpp/farm_ng/rs2_bag_to_event_log.cpp [hunk 8]
```diff
@@ -65,79 +67,91 @@ class Rs2BagToEventLogProgram {
...
+    camera_model.set_frame_name(configuration_.camera_frame_name());
+    SetCameraModelFromRs(camera_model, profile.get_intrinsics());
+
     Image image_pb;
     image_pb.mutable_fps()->set_value(30);
+    image_pb.mutable_camera_model()->CopyFrom(*camera_model);
     image_pb.mutable_frame_number()->set_value(0);
 
-    LoggingStatus log = StartLogging(bus_, configuration_.name());
+    encoder = " x264enc ! ";
```
ethanrublee, line 79: Suggest reusing VideoStreamer.

bootstrap-venv.sh
```diff
@@ -24,3 +24,6 @@ mkdir -p $FARM_NG_GOPATH
 export GOPATH=$FARM_NG_GOPATH:$GOPATH
 go get -u github.com/golang/protobuf/protoc-gen-go
 go get -u github.com/twitchtv/twirp/protoc-gen-twirp
+
+pip3 install wheel
```
isherman, line 28: let's remove this, seems out of scope for this change and @ethanrublee is actively updating the bootstrap/dev environment workflows right now

docker/dev.Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM ubuntu:18.04
```
isherman, line 1: let's remove this file, seems out of scope for this change and @ethanrublee is actively updating the bootstrap workflows right now

modules/perception/cpp/farm_ng/CMakeLists.txt
```diff
@@ -6,9 +6,15 @@ target_link_libraries(capture_video_dataset farm_ng_perception)
...
 target_link_libraries(create_video_dataset farm_ng_perception)
 
-
 add_executable(detect_apriltags detect_apriltags.cpp)
 target_link_libraries(detect_apriltags farm_ng_perception)
 
 add_executable(camera_pipeline camera_pipeline.cpp)
 target_link_libraries(camera_pipeline farm_ng_perception)
+
+# TODO: Should realsense2 symbols be exported from the perception library?
```
isherman, line 15: @ethanrublee do you like this as is?

modules/perception/cpp/farm_ng/rs2_bag_to_event_log.cpp
```diff
@@ -0,0 +1,237 @@
+/* Project docs and todo list:
+ *   https://docs.google.com/document/d/1cvZWOBCO7R57tfYvI6r97d-aeuYCi-Os10OqYn9Duqo/edit?usp=sharing
```
isherman, line 2: Let's avoid putting a link to a Google Doc into the source. Then we have to deal with permissions, etc. 
Right place for TODOs would be a Github issue (fine to link to). Not sure what you have for "project docs".

modules/perception/cpp/farm_ng/rs2_bag_to_event_log.cpp
```diff
@@ -0,0 +1,236 @@
...
+ * This program takes the path of an RS2 rosbag file, relative to the blobstore
+ * root, and outputs an mp4 file and an event log with the following folder
+ * structure:
+ * <Blobstore root>
+ *   - logs/<name>
+ *     - <camera_frame_name>-<timestamp>.mp4
+ *       - The program outputs mp4 files in 300-frame
+ *          chunks, so there may be multiple of these
+ *          files for each program run.
+ *     - events-<timestamp>.log
```
isherman, line 13: I updated the documentation a little bit. 
FYI, the numbers you see in the filename aren't a timestamp, but rather a PID (process id) and a counter (to avoid conflicting filenames). Should probably update these docs to reflect that.

modules/perception/cpp/farm_ng/rs2_bag_to_event_log.cpp
```diff
@@ -0,0 +1,236 @@
...
+namespace farm_ng {
+namespace perception {
+
+class Rs2BagToEventLogProgram {
+ public:
+  Rs2BagToEventLogProgram(EventBus& bus,
+                          Rs2BagToEventLogConfiguration configuration,
+                          bool interactive)
+      : bus_(bus), timer_(bus_.get_io_service()) {
+    if (interactive) {  // The program doesn't use interactive mode at this
```
isherman, line 62: can remove this comment (it's likely to go stale)

modules/perception/cpp/farm_ng/rs2_bag_to_event_log.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+    Rs2BagToEventLogResult result;
+
+    result.mutable_stamp_begin()->CopyFrom(MakeTimestampNow());
+
+    // Start a realsense pipeline from a recorded file to get the framesets
+    rs2::pipeline pipe;
+    rs2::config cfg;
+
+    // Set the optional repeat_playback arg to false
```
isherman, line 89: Put this comment inline with the argument for better legibility, less likely to go stale:
```
cfg.enable_device_from_file(
   (farm_ng::core::GetBlobstoreRoot() / configuration_.rs2_bag_resource().path()).string(),
    false // repeat_playback
);
```

Or some would prefer:
```
bool repeat_playback = false;
cfg.enable_device...(..., repeat_playback)

```

modules/perception/cpp/farm_ng/rs2_bag_to_event_log.cpp
```diff
@@ -0,0 +1,236 @@
...
+    camera_model.set_frame_name(configuration_.camera_frame_name());
+    SetCameraModelFromRs(&camera_model, color_stream.get_intrinsics());
+
+    // This variable will be populated by the video streamer. Declare
+    // outside of the while loop to access for use by result message.
+    Image image_pb;
+
+    std::unique_ptr<VideoStreamer> streamer = std::make_unique<VideoStreamer>(
+        bus_, camera_model,
+        // later we will replace MODE_MP4_FILE with MODE_JPG_SEQUENCE
```
isherman, line 109: Maybe just replace with `TODO(collinbrake): Support MODE_JPG_SEQUENCE`

modules/perception/cpp/farm_ng/rs2_bag_to_event_log.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+    CameraModel camera_model;
+    camera_model.set_frame_name(configuration_.camera_frame_name());
+    SetCameraModelFromRs(&camera_model, color_stream.get_intrinsics());
+
+    // This variable will be populated by the video streamer. Declare
+    // outside of the while loop to access for use by result message.
+    Image image_pb;
+
+    std::unique_ptr<VideoStreamer> streamer = std::make_unique<VideoStreamer>(
```
isherman, line 107: Prefer to declare this on the stack, not the heap, as its lifecycle is the same as its scope (and internally it will do any heap allocation it needs).
@ethanrublee can correct me if I'm wrong.

modules/perception/cpp/farm_ng/rs2_bag_to_event_log.cpp
```diff
@@ -0,0 +1,236 @@
...
+      cv::Mat color = RS2FrameToMat(color_frame);
+      // cv::Mat depth = RS2FrameToMat(depth_frame),
+
+      if (color.empty()) {
+        break;
+      }
+
+      // Display the current video frame, using case insensitive q as quit
+      // signal.
+      cv::imshow("Color", color);
```
isherman, line 142: This is more of debugging code, probably best not to check it in.

modules/perception/cpp/farm_ng/rs2_bag_to_event_log.cpp
```diff
@@ -0,0 +1,236 @@
...
+        break;
+      }
+
+      // Add timestamped frame with the video streamer
+      auto frame_stamp =
+          google::protobuf::util::TimeUtil::MillisecondsToTimestamp(
+              color_frame.get_timestamp());
+      image_pb = streamer->AddFrame(color, frame_stamp);
+
+      // std::cout << "Wrote frame number: " << image_pb.frame_number().value()
```
isherman, line 155: Remove

modules/perception/cpp/farm_ng/rs2_bag_to_event_log.cpp
```diff
@@ -0,0 +1,236 @@
...
+      bus_.Send(
+          MakeEvent(camera_model.frame_name() + "/image", image_pb, stamp));
+
+      // zero index base for the frame_number, set after send.
+      image_pb.mutable_frame_number()->set_value(
+          image_pb.frame_number().value() + 1);
+    }
+
+    result.mutable_configuration()->CopyFrom(configuration_);
+    result.mutable_dataset()->CopyFrom(image_pb.resource());
```
isherman, line 169: This should point to the event log, not the (latest chunked) video file. See `create_video_dataset` for reference.

modules/perception/cpp/farm_ng/rs2_bag_to_event_log.cpp
```diff
@@ -0,0 +1,236 @@
...
+    result.mutable_stamp_end()->CopyFrom(MakeTimestampNow());
+
+    ArchiveProtobufAsJsonResource(configuration_.name(), result);
+
+    LOG(INFO) << "Complete:\n" << status_.DebugString();
+
+    status_.set_num_frames(image_pb.frame_number().value());
```
isherman, line 176: Move this inside the while loop (it will be updated at every frame, and sent on the event bus once per second).

modules/perception/cpp/farm_ng/rs2_bag_to_event_log.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+}  // namespace perception
+}  // namespace farm_ng
+
+int Main(farm_ng::core::EventBus& bus) {
+  farm_ng::perception::Rs2BagToEventLogConfiguration config;
+  config.set_name(FLAGS_name);
+  config.set_camera_frame_name(FLAGS_camera_frame_name);
+  config.mutable_rs2_bag_resource()->set_path(FLAGS_rs2_bag_path);
+  config.mutable_rs2_bag_resource()->set_content_type("rosBag/rs2");
```
isherman, line 222: let's use `application/x-rosbag` instead. Read up on Mime Types if you're curious about the conventions.

modules/perception/protos/farm_ng/perception/rs2_bag_to_event_log.proto
```diff
@@ -0,0 +1,46 @@
...
+import "farm_ng/core/resource.proto";
+//import "farm_ng/perception/capture_video_dataset.proto";
+
+import "google/protobuf/timestamp.proto";
+import "google/protobuf/wrappers.proto";
+
+package farm_ng.perception;
+option go_package = "github.com/farm-ng/genproto/perception";
+
+message Rs2BagToEventLogConfiguration {
```
isherman, line 12: By convention, our programs (and their associated protos) start with a verb. "Capture" a dataset. "Calibrate" a thing. "Detect" apriltags. Let's name this program "convert_rs2_bag". The proto file, and proto messages should be updated to reflect that change as well.

A bit annoying to change (sorry), but I think it's a good convention to maintain.

modules/perception/protos/farm_ng/perception/rs2_bag_to_event_log.proto
```diff
@@ -0,0 +1,46 @@
+syntax = "proto3";
+
+import "farm_ng/core/resource.proto";
+//import "farm_ng/perception/capture_video_dataset.proto";
```
isherman, line 4: Remove

modules/perception/protos/farm_ng/perception/rs2_bag_to_event_log.proto
```diff
@@ -0,0 +1,46 @@
...
+message Rs2BagToEventLogConfiguration {
+
+  // Name of the resulting resource
+  string name = 1;
+
+  string camera_frame_name = 2;
+
+  // This resource will store the path to the input
+  // bag file.
+  farm_ng.core.Resource rs2_bag_resource = 3;
```
isherman, line 21: Remove the `_resource` suffix; the type tells us it's a resource.

modules/perception/protos/farm_ng/perception/rs2_bag_to_event_log.proto
```diff
@@ -0,0 +1,46 @@
...
+}
+
+
+message Rs2BagToEventLogStatus {
+  oneof input_required {
+    Rs2BagToEventLogConfiguration input_required_configuration = 1;
+  }
+
+  // The number of bag file frames exported
```
isherman, line 30: Throughout, let's use "converted / conversion" rather than "exported".

modules/perception/protos/farm_ng/perception/rs2_bag_to_event_log.proto
```diff
@@ -0,0 +1,46 @@
...
+  int32 num_frames = 2;
+}
+
+message Rs2BagToEventLogResult {
+  // Configuration which initiated the video/log export
+  Rs2BagToEventLogConfiguration configuration = 1;
+
+  // The time the capture began
```
isherman, line 38: "conversion"

modules/perception/cpp/farm_ng/CMakeLists.txt
ethanrublee: Here we can remove the find_package, just check if realsense2_found, and the realsense2_LIBRARY linkage is likely unnecessary, but it doesn't hurt.

modules/perception/cpp/farm_ng/rs2_bag_to_event_log.cpp
ethanrublee: Since you have the camera_model, yeah, stack would be great.

modules/perception/protos/farm_ng/perception/rs2_bag_to_event_log.proto
ethanrublee: Oh, i didn't realize the convention sorry!  But i like it :)

modules/perception/cpp/farm_ng/CMakeLists.txt
isherman: 👍 . We found we did need to link on realsense2, as it's not part of the perception library's public link interface.

isherman (approved):

modules/perception/cpp/farm_ng/CMakeLists.txt
```diff
@@ -6,9 +6,14 @@ target_link_libraries(capture_video_dataset farm_ng_perception)
...
 target_link_libraries(create_video_dataset farm_ng_perception)
 
-
 add_executable(detect_apriltags detect_apriltags.cpp)
 target_link_libraries(detect_apriltags farm_ng_perception)
 
 add_executable(camera_pipeline camera_pipeline.cpp)
 target_link_libraries(camera_pipeline farm_ng_perception)
+
+find_package(realsense2)
```
isherman, line 15: Thought I saw this removed in one of your commits, but looks like it came back

collinbrake:
@isherman the process ID on the event log is not the same as the PID for the video files and the JSON file. For example:
image
image

isherman:
> @isherman the process ID on the event log is not the same as the PID for the video files and the JSON file. For example:

Makes sense. Remember the events log is created by the ipc-logger process (we just tell it via the event bus to `StartLogging`). The video files and result file are created by the program itself.

modules/perception/protos/farm_ng/perception/camera_model.proto
```diff
@@ -10,14 +10,18 @@ message CameraModel {
   int32 image_width = 1;
   // Image height in pixels.
   int32 image_height = 2;
+  // Depthmap width in pixels.
+  int32 depthmap_width = 3;
```
isherman, line 14: I would avoid renumbering fields in existing protobuf definitions, since we already have datasets captured with the existing field numbers. Keeping the field numbers consistent is what allows forwards/backwards compatibility as the message schema evolves: https://developers.google.com/protocol-buffers/docs/proto#updating

It's common to see "out of order" numbers in a protobuf definition for this reason.

isherman: oops, should be proto3 link: https://developers.google.com/protocol-buffers/docs/proto3#updating

modules/perception/cpp/farm_ng/convert_rs2_bag.cpp
```diff
@@ -0,0 +1,239 @@
...
+    Image image_pb;  // result must have access to this
+                     // outside of the loop
+
+    VideoStreamer streamer =
+        VideoStreamer(bus_, camera_model,
+
+                      // TODO(collinbrake): Support MODE_JPG_SEQUENCE
+                      VideoStreamer::MODE_MP4_FILE);
+
+    LoggingStatus log = StartLogging(bus_, configuration_.name());
```
ethanrublee, line 115: Since we don't support namespacing yet in the IPC log, and I want to run multiple bags at the same time, I think this should just use the EventLogger class and no IPC.

modules/perception/cpp/farm_ng/convert_rs2_bag.cpp
```diff
@@ -0,0 +1,239 @@
...
+        const int wd = depth_frame.get_width();
+        const int hd = depth_frame.get_height();
+
+        CHECK_EQ(camera_model.image_width(), wc);
+        CHECK_EQ(camera_model.image_height(), hc);
+
+        // Image matrices from rs2 frames
+        cv::Mat color = RS2FrameToMat(color_frame);
+        cv::Mat depth;
+        RS2FrameToMat(depth_frame).convertTo(depth, CV_8UC1);
```
ethanrublee, line 138: What is the type of the depth mat before the convertTo?

modules/perception/cpp/farm_ng/perception/video_streamer.cpp
```diff
@@ -48,12 +48,23 @@ void VideoStreamer::ResetVideoStreamer(bool is_color) {
...
+        image_pb_.camera_model().frame_name() + "/color", "mp4", "video/mp4");
     gst_pipeline =
         std::string("appsrc !") + " videoconvert ! " + encoder +
         " mp4mux ! filesink location=" + resource_path.second.string();
     image_pb_.mutable_resource()->CopyFrom(resource_path.first);
     image_pb_.mutable_frame_number()->set_value(0);
+
+    // Populate the Depthmap field of Image here and create
+    // gstreamer pipeline for depth map writing
+    auto depthmap_path = GetUniqueArchiveResource(
```
ethanrublee, line 60: This should likely be only constructed if depth maps are expected.

modules/perception/protos/farm_ng/perception/camera_model.proto
```diff
@@ -10,6 +10,10 @@ message CameraModel {
   int32 image_width = 1;
   // Image height in pixels.
   int32 image_height = 2;
+  // Depthmap width in pixels.
```
ethanrublee, line 13: Huh, the current expectation is that the depthmap is mapped into the color image frame, if this is not the case, then we need to track more than width/height, the depthmap will have its own camera model, and extrinsic relationship to the visual imager.

ethanrublee: Suggest removing these fields, and align the depth image into the resolution and reference frame of the color image.

See this sample:
https://github.com/IntelRealSense/librealsense/tree/master/examples/align

ethanrublee: ``rs2::align align_to_color(RS2_STREAM_COLOR);``

modules/perception/cpp/farm_ng/perception/video_streamer.cpp
```diff
@@ -98,8 +116,38 @@ Image VideoStreamer::AddFrame(const cv::Mat& image,
   }
   return image_sent;
 }
+Image VideoStreamer::AddFrameWithDepthmap(
```
ethanrublee, line 119: Here we should document and CHECK the type expectation for depthmap.

Suggest that it can be either  ``CV_16UC1 // millimeters``  or ``CV_32FC1 //meters``

ethanrublee (changes requested):
Let's wait until this is more feature complete.

- Support image sequences (jpeg)
- Remap the depth image to the color image
- Save out depth stream as linear mm 16 bit png sequence at least
- Don't use IPC logger for now.

modules/perception/cpp/farm_ng/perception/video_streamer.h
```diff
@@ -17,9 +17,16 @@ class VideoStreamer {
...
-    MODE_MP4_UDP = 3
+    MODE_JPG_SEQUENCE = 2,
+    MODE_MP4_FILE = 3,
+    MODE_MP4_UDP = 4,
+    MODE_DEPTH = 5
   };
+  enum DepthMode {
+    DEPTH_MODE_UNDEFINED = 0,
+    DEPTH_MODE_LINEAR_16BIT_PNG = 1,
+    DEPTH_MODE_LINEAR_8BIT_JPG = 2
```
ethanrublee, line 28: LINEAR 8 bit doesn't make sense due to the range considerations.  This should be INVERSE

collinbrake:
- call AddFrame from AddFrameDepth
- populate name from file base name
- examples for image writer:
  - `multiview_april_tag_rig_calibrater()`

modules/perception/cpp/farm_ng/perception/video_streamer.cpp
```diff
@@ -21,7 +21,7 @@ VideoStreamer::VideoStreamer(EventBus& bus, const CameraModel& camera_model,
 
 void VideoStreamer::ResetVideoStreamer(bool is_color) {
   // TODO(ethanrublee) look up image size from realsense profile.
-  std::string gst_pipeline;
+  std::string gst_pipeline, gst_pipeline_depthmap;
```
ethanrublee, line 24: TODO i think these changes can be reverted.

modules/perception/cpp/farm_ng/perception/video_streamer.h
```diff
@@ -17,24 +17,37 @@ class VideoStreamer {
   enum Mode {
     MODE_UNDEFINED = 0,
     MODE_PNG_SEQUENCE = 1,
-    MODE_MP4_FILE = 2,
-    MODE_MP4_UDP = 3
+    MODE_JPG_SEQUENCE = 2,
```
ethanrublee, line 20: I think changes to video_streamer.h can be reverted for now, as this implementation is now in the bag conversion program.

collinbrake:
@ethanrublee Do you want the default dataset name to be the base filename of the bag file instead of "default"?

PR #185, opened 2020-12-28
Sketch using bootstrap-apt and bootstrap-venv from devel.Dockerfile, …

ethanrublee:
…prepping for using docker as main develeopment environment on x86 and l4t


isherman:
LGTM so far

PR #192, opened 2021-01-16
Improve devel + pr ci times.

ethanrublee:
- Use ccache, locally and in github action
- Use action cache to skip building devel docker files if nothing changed
- Use the devel image for CI rather than base.


ethanrublee:
Going to merge - now CI builds on prs with a warm cache take less than 5 minutes. And updates to the devel container are similarly fast.

PR #215, opened 2021-03-30
Display program stdout and stderr in frontend

isherman:
A first pass at piping stdout and stderr from programd launched programs to the frontend.

Obviously the way this data is displayed is not great; but trying to find a simple solution that avoids a major rethinking of the programs UI. 

Output could be batched into N-second chunks instead of sent line-by-line. 

Not a solution for very high-volume output streams (e.g. https://github.blog/2021-03-25-how-github-actions-renders-large-scale-logs/)


modules/core/python/farm_ng/core/programd.py
```diff
@@ -87,12 +88,31 @@ async def launch_child_process(self, program_info):
...
 
+    def io_stream_to_event_bus(self, stream, out_buffer):
+        while True:
+            line = yield from stream.readline()
+            if not line:
+                return
+            out_buffer.write(line)
+            out_buffer.flush()
+            event = make_event(
+                f'{self.status.running.program.id}/status',
```
ethanrublee, line 108: i think here status is over loaded with more programatic {id}/status messages, suggest using a different name such as {id}/output, {id}/{stdout|stderr}

modules/frontend/frontend/src/components/scope/visualizers/DefaultProgramStatusVisualizer.tsx
```diff
@@ -0,0 +1,31 @@
...
+  } = props;
+
+  return (
+    <Card timestamp={timestamp} json={value}>
+      {value.line}
+    </Card>
+  );
+};
+
+export const DefaultProgramStatusVisualizer = {
```
ethanrublee, line 25: I feel like the program output should be available in addition to status.  As developers tend to add debug output that's useful during corner cases that the status visualization may not help communicate.

isherman: ok, we can go that route. actually simplifies things to not try to shoehorn it into a status visualizer.

isherman: done

isherman:
Pushed some additional commits that render stdout and stderr, merged into a single UI component:

Screen Shot 2021-04-01 at 1 59 32 AM

Still does not scale well to long logs. 

PR also now includes some cleanup of duplicated code.

ethanrublee:
Thank you!

ethanrublee (approved):

ethanrublee:
So in testing this I've noticed an interesting feedback case.  The program being run subscribes to its ``{program_name}/*``, and then for every line it prints to the log, it receives a message.  If on receipt it prints something to the log, then things tend to explode.

I suggest we change this to be programd/stdout, programd/stdin, as at the moment programd only runs one program at a time.

isherman:
> So in testing this I've noticed an interesting feedback case. The program being run subscribes to its `{program_name}/*`, and then for every line it prints to the log, it receives a message. If on receipt it prints something to the log, then things tend to explode.
> 
> I suggest we change this to be programd/stdout, programd/stdin, as at the moment programd only runs one program at a time.

Oh, good catch. I like that idea -- is that a change you'd like me to make or do you plan to make it?

ethanrublee:
I'm trying to make it now, will let you know if I get stuck.

PR #217, opened 2021-04-06
Rename ProgramOutput messages to avoid self subscription feedback.

ethanrublee:
Would be nice to add auto scroll somehow to follow the output.


isherman (approved):
Looks good. Happy to take a look at auto-scroll when I have a chance.

PR #226, opened 2021-05-25
Support for loading grouped intrinsic calibration files when generating datasets

greidy:


modules/perception/cpp/farm_ng/perception/create_video_dataset_program.cpp
```diff
@@ -50,6 +51,16 @@ CreateVideoDatasetProgram::CreateVideoDatasetProgram(
   bus_.GetEventSignal()->connect(std::bind(&CreateVideoDatasetProgram::on_event,
                                            this, std::placeholders::_1));
   on_timer(boost::system::error_code());
+
+  // initialize camera calibration db
+  for (const farm_ng::core::Resource &r : configuration_.group_intrinsic_calibration())
```
ethanrublee, line 56: suggest not using single letter names, except for i/j/k...

modules/perception/cpp/farm_ng/perception/image_loader.cpp
```diff
@@ -18,17 +18,28 @@ void ImageLoader::OpenVideo(const Resource& resource) {
...
-    if (video_name_ != (GetBlobstoreRoot() / resource.path()).string()) {
+    if (video_name_ != AbsolutePathToResource(resource.path())) {
       capture_.reset(nullptr);
       OpenVideo(resource);
     }
   }
   CHECK(capture_->isOpened()) << "Video is not opened: " << video_name_;
 }
 
+std::string ImageLoader::AbsolutePathToResource(const std::string& relativePath) {
```
ethanrublee, line 32: I think there is a function like this in ``core/blobstore.h``?  Not sure if it would work for your use case: https://github.com/farm-ng/tractor/blob/master/modules/core/cpp/farm_ng/core/blobstore.cpp#L27

ethanrublee (changes requested):
Some minor nits.  Sorry for the delay in review!

modules/calibration/protos/farm_ng/calibration/calibrate_intrinsics.proto
```diff
@@ -49,3 +49,10 @@ message CalibrateIntrinsicsResult {
   google.protobuf.Timestamp stamp_begin = 6;
   google.protobuf.Timestamp stamp_end = 7;
 }
+
+message GroupCalibrateIntrinsicsResult {
```
ethanrublee, line 53: Maybe this should be ``MultiView`` or ``MultiViewRig`` to be consistent with the use of multi_view in the extrinsic solver.

modules/perception/cpp/farm_ng/perception/image_loader.cpp
greidy: Yes indeed - fixed.

modules/calibration/protos/farm_ng/calibration/calibrate_intrinsics.proto
greidy: Deferred

ethanrublee (approved):

PR #244, opened 2021-07-30
Estimate the alignment given an apriltag rig.

ethanrublee:


modules/calibration/cpp/farm_ng/align_sensor_rig.cpp
```diff
@@ -207,5 +350,5 @@ int main(int argc, char* argv[]) {
     farm_ng::core::WriteProtobufToJsonFile(FLAGS_output_config, config);
     return 0;
   }
-  return farm_ng::calibration::align_sensor_rig(config);
+  return farm_ng::calibration::align_sensor_rig_apriltag_rig(config);
```
ethanrublee, line 353: I guess this should be a switch in case we want to revert to the other method using a few hand measured points and the ground plane.

